PR #131, opened 2026-08-05
fix: restore tmux attach terminals concurrently after window reload

hzcheng:
## Summary

Window reload on a workspace with many persistent terminals stalled tmux session restoration for ~22s (observed in production telemetry on a window with ~24 terminals; earlier single-window reloads took 5-7s). The dashboard itself stayed fast (first paint 1.2s, bootstrap ready 2.2s — the 800ms deferral budget worked as designed), but AI session cards kept their pre-restore state until the background attach restore settled.

Root cause: `restoreAttachTerminalsOnce` awaited `terminal.processId` **serially** for every terminal — including plain non-tmux terminals — with a per-terminal 2s timeout, right while the pty host was still reconnecting the persisted processes. It also issued one `tmux list-clients` invocation per terminal for live-client lookups.

## Changes

- `tmuxRuntimeBackend.restoreAttachTerminalsOnce`: resolve all untracked terminal process IDs **concurrently** (`Promise.all`, per-terminal timeout preserved), then process them serially as before. Live tmux client lookups now share one lazily-fetched `list-clients` snapshot per restore pass.
- `TmuxClient`: new `getClientSessionsByProcess()` returning the full `pid → session` map; `getClientSessionForProcess()` keeps its validation and delegates (no behavior change).
- No behavior change: manual `tmux attach` terminals created outside the extension remain restorable (that is exactly why the process-ID wait is kept for non-candidate terminals, just parallelized).

Expected effect on the observed scenario: ~22s → ~2-4s (bounded by pty host reconnection, not by our own serialization). Steady-state impact: none (this path runs once per window boot/reload).

## Tests and contracts

- New contract `RUNTIME-TMUX-ATTACH-RESTORE-CONCURRENCY-001` (P1, automated), owned by `tests/contract/aiSessions/runtimeBackends.test.js`:
  - restore subscribes to every terminal process ID before any of them settles (deterministic blocking-thenable pin, no timers);
  - one shared live-client list per restore pass, zero per-terminal `list-clients` calls;
  - terminals whose process ID never resolves are skipped without blocking the rest.
- `tmuxClientBehavior.test.js`: batched list-clients snapshot (map contents, no-server → empty map, invalid output, privacy-safe nonzero-exit error).
- Mutation-checked: all three new tests fail against the previous serial implementation.
- Hand-rolled tmux client fakes updated for the new interface (`tests/helpers/runtimeContract.js`, `scripts/run-ai-session-tmux-checks.js`).

## Gates

- Full local gates green (compile, unit, integration, dashboard, safety, guards, baseline, browser, contracts, lint, packaging).
- Changed-line coverage: **100.00% (24/24)** against `origin/main`.
- Capability audit: commits assigned to `MAIN-RUNTIME-SESSION-RECOVERY`, behavior registered, audit commit `8c58898`.

## Skill harvest

none — the only process friction was a hand-rolled tmux client fake missing the new interface method; the safety gate caught it deterministically in 31s with a self-explanatory `TypeError`, so no reusable instruction gap exists.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,15 @@ All notable changes to the "Agent Pivot" extension will be documented in this fi
   telemetry pill now reports open/total (e.g. `Comments 2/5`). Existing stored
   comments migrate automatically (sent and resolved become done).
 
+### Fixed
+
+- Window reloads no longer stall tmux session restoration on windows with many
+  terminals: attach restore now resolves terminal process IDs concurrently
+  (previously serial with a per-terminal timeout, which multiplied the delay
+  by the terminal count while the pty host was still reconnecting) and shares
+  a single live-client list across all terminals instead of one tmux
+  invocation per terminal.
+
 ## [1.0.3] - 2026-08-04
 
 ### Added
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -1646,6 +1646,19 @@
     ]
   },
   {
+    "id": "RUNTIME-TMUX-ATTACH-RESTORE-CONCURRENCY-001",
+    "domain": "runtime",
+    "title": "Reload attach restore resolves terminal process IDs concurrently and shares one live-client list per pass",
+    "priority": "P1",
+    "status": "automated",
+    "owners": [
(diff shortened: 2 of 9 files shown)


hzcheng:
lgtm

PR #132, opened 2026-08-05
fix: exempt the gate-building merges from the approval audit

hzcheng:
## Summary

The push-only `merge-approval-audit` job turned main red after #131: it flagged #126 and #128 — the two PRs that **built** the approval gate — for lacking an owner approval comment. Those merges (2026-08-05T01:57:01Z / 02:05:43Z) happened under the previous in-conversation approval rules and by definition could not carry the comment the gate introduced.

Root cause: `MERGE_APPROVAL_REQUIRED_SINCE` was anchored to the calendar boundary `2026-08-05T00:00:00Z` instead of to the last pre-gate merge. This fix moves it to `2026-08-05T02:06:00Z` (just after #128) with a comment recording the anchoring rule.

## Verification

- Reproduced the failure locally, then ran the audit script against the live repository with the fix: `#131/#130/#129 approval marker present`, `#126/#128` correctly exempt — `rc=0`.
- Existing cutoff unit test pins the boundary symbolically (`mergeRequiresApproval(MERGE_APPROVAL_REQUIRED_SINCE) === true`), so it survives the bump unchanged.
- Full local gates green; changed-line coverage 100% (5/5).

## Skill harvest

none — the lesson (anchor activation cutoffs to the last exempt event, not a calendar boundary) is recorded in the code comment at the site where a future rollout would copy the pattern; no reusable skill instruction gap beyond that.

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "fb76fbaed8adbae010148a9d0322e7c9f3a40a9f",
+        "head": "47312e9a48be6aacebd147e4cb717727e49c0b2e",
         "ignoredDocumentationCommits": [
             "dd86a325e3db5aa013ffa18a647e67e9fa279c10",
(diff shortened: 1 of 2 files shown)


hzcheng:
lgtm

PR #137, opened 2026-08-05
Fix repeated AI Conversation selection highlights

hzcheng:
## Summary

- Highlight only the selected user input when navigating to a conversation interaction.
- Do not replay the temporary selection highlight during live refreshes.
- Cover initial navigation and repeated refresh behavior in Chromium.

## Verification

- `npm run test-compile`
- `node --test --test-concurrency=1 tests/browser/conversationViewer.test.js`
- `npm run lint`
- `git diff --check`

## Skill harvest

- `none`: the task exposed no reusable workflow gap beyond existing repository skills and process rules.

No version release is included.


hzcheng:
lgtm

PR #139, opened 2026-08-05
feat: switch active sessions from floating viewer rails

hzcheng:
## Summary

- Add two large floating rails to the AI Conversation viewer — bottom-left **Previous active session** and bottom-right **Next active session** — that cycle the viewer through the current workspace's active sessions in dashboard (Active list) order, with wrap-around.
- Ghost design: compact 36px circular buttons resting at 30% opacity with a soft shadow, small enough that corner text stays selectable for commenting; on hover or keyboard focus they become fully opaque, lift 2px, deepen the shadow, and pick up the theme focus border. Chevron-with-bar glyphs stay visually distinct from the in-conversation ‹ › header navigation. Honors `prefers-reduced-motion` and `forced-colors`.
- The rails sit in a pointer-transparent grid overlay pinned to the conversation column, so they track the column edge when the side panel opens or is resized (never covering the comments toolbar) and never block text selection. The scroll surface gains 68px bottom clearance so the last line never rests underneath a rail.

## Behavior and protocol

- The webview only posts a direction; the Host supplies the current target from its own state and resolves the neighbor from the authoritative ordered active list (pending sessions without a conversation are skipped), anchors at the neighbor's latest interaction, and swaps the document in place via the existing follow path.
- New `conversation-viewer-switch-session` v1 message (`direction: 'previous' | 'next'`), validated fail-closed with exact keys. Newest intent wins via the shared viewer intent generation (a slow switch is superseded by a newer one).
- With fewer than two switchable sessions a switch is a deliberate no-op (`noAdjacentSession`), consistent with how in-conversation Previous/Next behave at boundaries.
- A successful switch also focuses the adjacent session's terminal/tmux window (`focusSession` → `focusActive`), mirroring sidebar card activation so the visible runtime always matches the conversation on screen. Terminal sync is best-effort: a focus failure never undoes a settled switch.

## Tests

- Unit: protocol accept/reject samples for the new message.
- Integration (viewer host): routes exact switches with the authoritative current target; malformed/over-posted intents stay inert.
- Integration (composition): dashboard-order adjacency, wrap-around, pending-session skip, `noAdjacentSession` / `closed` / `unavailable` results, and newest-switch-wins supersede.
- Browser: ghost computed styles (36px circle at 30% opacity), hover promotion to full opacity, corner geometry against the viewport, side-panel containment, and exact posted intents. Pre-existing `getByRole` name queries for the header Previous/Next/Latest buttons were pinned with `exact: true` because the new rails' accessible names legitimately contain those words.
- Verified end to end with `npm run test:ci:linux` (green, changed-line coverage 96.75%) and rendered screenshots (idle ghost, hover, side panel open).

## Skill harvest

none — the local RED-GREEN loop plus a full `test:ci:linux` run caught every issue before push (aria-name selector collision, hover transition timing); no workflow friction or user correction justified a skill change.


hzcheng:
lgtm

PR #142, opened 2026-08-06
Fix local file links in AI Conversation

hzcheng:
## Summary

- render absolute local file references in AI Conversation as clickable links
- open file links in the workspace Extension Host at the requested line and column
- preserve the existing sanitizer boundary for unsafe schemes and relative URLs

## Skill harvest

no skill change — the task corrections were compliance gaps already covered by the regression and worktree skills

## Verification

- `npm run test:ci:linux`
- `npm run test:behavior-contracts` after the capability audit commit
- packaged and installed `hzcheng.agent-pivot@1.0.4` in the active remote VS Code Server with manifest and runtime hash verification
- user manually verified the installed build

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -3909,6 +3909,24 @@
     ]
   },
   {
+    "id": "CONVERSATION-LOCAL-FILE-LINKS-001",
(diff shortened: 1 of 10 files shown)


hzcheng:
lgtm

PR #144, opened 2026-08-06
Fix sidebar startup rendering stalls

hzcheng:
## Summary

- render the authoritative sidebar document without waiting for terminal/runtime recovery
- restore tmux and Direct runtimes independently, publishing each result as it settles
- gate only startup storage mutations so lazy panel hydration stays responsive
- rebuild the deterministic dashboard bundle whenever watched webview assets change
- update tmux and dashboard safety harnesses for the production startup protocol

## Root cause

The ready document could render early, but deferred runtime recovery was still serialized: tmux waited for Direct process-ID restoration, and refresh publication could wait behind visibility preparation. A broad migration barrier could also delay read-only panel hydration. Separately, the CI tmux fake still expected individual metadata calls while production now batches them.

## Verification

- `npm run test:ci:linux`
- `AGENT_PIVOT_TMUX_PATH=/usr/bin/tmux npm run test:tmux:smoke`
- focused runtime composition tests
- `git diff --check`

Changed-line coverage: 93.29% (278/298).

## Skill harvest

No skill update was justified. The existing regression and PR workflows already require reproducing the exact failing gate and rerunning fresh complete verification; these were product and harness regressions covered by new tests, not an instruction gap.

## Release

No version is published by this PR.


hzcheng:
lgtm

PR #148, opened 2026-08-06
Fix sidebar initial render starvation

hzcheng:
## Summary

- keep the current Dashboard Webview document alive while it is still loading and coalesce startup refreshes into one latest render
- correlate renderer-ready acknowledgements with a document generation so stale acknowledgements cannot release a newer document
- recover a genuinely missing acknowledgement with one delayed retry, without creating a periodic reload storm
- cover the Host lifecycle and the production Dashboard HTML/CSS/bundle handshake in CI

## Skill harvest

no skill change — the false start was a product-specific recovery-policy defect; the existing regression workflow already requires RED evidence and production rendered-surface ownership, which this branch now supplies.

## Verification

- `npm run test:ci:linux`
  - deterministic: 331/331
  - browser: 135/135
  - changed-line coverage: 98.58% (139/141)
  - behavior contracts, safety, architecture, release packaging: passed
- final read-only integration review: no Critical or Important findings
  - installed `hzcheng.agent-pivot@1.0.4`
  - installed bytes match the final worktree build
- manual reload validation in two remote VS Code windows

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -2231,7 +2231,9 @@
     "priority": "P0",
     "status": "automated",
     "owners": [
-      "tests/integration/dashboard/errorRecovery.test.js"
+      "tests/browser/dashboardStylesheetStartup.test.js",
+      "tests/integration/dashboard/errorRecovery.test.js",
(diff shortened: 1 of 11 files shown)


hzcheng:
lgtm

PR #150, opened 2026-08-06
Fix Claude context window telemetry

hzcheng:
## Summary

- derive Claude context-window capacity from the resolved model instead of a global 200k fallback
- cover the current 1M model matrix, provider-prefixed IDs, Mythos Preview, and legacy 200k behavior
- verify the adapter against a real local Claude Opus 5 transcript without exposing conversation content

## Skill harvest

No skill change was justified. The worktree dependency-install retry was already covered by the existing worktree skill, and the installation workflow already required artifact and installed-byte verification.

## Verification

- npm run test-compile
- node --test tests/contract/aiSessions/claudeConversationAdapter.test.js
- npm run test:ci:linux
- git diff --check
- local VSIX installation with packaged-to-installed byte verification

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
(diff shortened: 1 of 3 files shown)


hzcheng:
lgtm

PR #151, opened 2026-08-06
Cut sidebar startup latency for Open Windows

hzcheng:
## Summary

Two measured startup latency fixes for the sidebar, plus the disproof of a third hypothesis so nobody retries it.

- **Start the Open Windows bridge handshake before bootstrap.** The bridge client was constructed 590-1055ms into bootstrap, and its constructor is what kicks off the handshake — a cross-host round trip to the UI extension host. That round trip therefore ran strictly *after* all local dashboard construction instead of alongside it. The client is now created at the top of `activate()` from a publication derived straight from VS Code state; aggregates, status and pin snapshots arriving before the dashboard controller exists are buffered latest-wins and replayed on adoption. The client outlives a bootstrap generation, so a retry re-adopts it instead of registering its commands twice.
- **Show Open Windows as `connecting` before the bridge answers.** `bridgeStatus` started at `ready`, so for seconds after every reload the section rendered a settled-looking "Live" empty list while the window data was still in flight. `unavailable` and `update-required` still force the section open; `connecting` respects an existing collapse preference rather than snapping shut a few seconds later.
- **Instrument `sinceModuleLoadMs`** on `activation-entered`, splitting the previously unattributed gap between activation dispatch and `activate()`.

## Measured effect

Production telemetry from the Agent Pivot output channel, 8 reloads on the new build vs 31 on the old, medians:

| | old (n=31) | new (n=8) | delta |
|---|---|---|---|
| bridge client created | 806ms | **4ms** | −802ms |
| handshake completed | 1789ms | 1204ms | −585ms |
| **first aggregate** | 2974ms | **2227ms** | **−747ms** |
| bootstrap ready (control) | 1039ms | 1069ms | +30ms (noise) |

The Open Windows section now fills in ~0.75s earlier, and `ready` is unchanged, so the gain is not paid for elsewhere.

Note the round trip itself got *slower* when started earlier (600-1300ms → 900-2027ms), which means a large part of the handshake is gated on UI-host readiness rather than transport. Starting earlier is now exhausted as a lever: the client is created at +4ms.

`sinceModuleLoadMs` measures 315-791ms — the gap between our module body finishing and VS Code calling `activate()`. That is extension-host contention, not our `require`, so shrinking the bundle or lazy-requiring cold paths is not a promising lever either.

## Disproved hypothesis (recorded so it is not retried)

An earlier revision of this branch added a bounded gate that held bootstrap until the two-stage boot skeleton was painted, on the hypothesis that `resolveWebviewView` was starved by our uninterrupted microtask drain (shell delay correlated with bootstrap duration at r=0.987 over 39 reloads). Production disproved it: the gate reported `painted=false` on 2/2 reloads, with the view resolve arriving up to 746ms *after* the 250ms budget expired, and `browser-first-paint` still absent. The view resolve is delivered on VS Code's own schedule; the correlation came from both timings sharing extension-host contention as a cause. The gate was reverted — it only added 250ms.

## Skill harvest

no skill change — all three lessons were compliance gaps against rules the existing skills already state, not missing guidance:

1. A `git rebase` ran against the primary checkout because the shell cwd had reset; `protecting-main-with-worktrees` already requires `git -C <worktree>`. No changes were made (the user's dirty files aborted it) and the primary checkout was verified unchanged.
2. A harness contract pass was reported as end-to-end proof that the paint gate worked in production; `fixing-regressions-with-ci` already stops on "a fake covers enough of a real environment". Production telemetry later disproved the claim.
3. The capability manifest was hand-edited, which is exactly why the rebase left stale commit hashes; `fixing-regressions-with-ci` already requires `regenerate-capability-audit.js`, which was used for the final audit commit.

## Verification

- `npm run test:ci:linux` — exit 0
  - deterministic: 612 unit / 719 contract / 336 integration
  - browser: 137
  - behavior contracts, safety, architecture, release packaging: passed
  - changed-line coverage: 100.00% (153/153) against `origin/main`
- rebased onto `origin/main` (`2344b9f`, includes #150) and re-run after the rebase
- built and installed locally via `npm run install-local` with VSIX-to-installed byte verification; the measurements above come from that installed build

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -1935,10 +1935,14 @@
     "priority": "P1",
     "status": "automated",
     "owners": [
-      "tests/contract/openProjects/bridgeClient.test.js"
+      "tests/contract/openProjects/bridgeClient.test.js",
+      "tests/contract/openProjects/dashboardConnectingStatus.test.js",
+      "tests/integration/dashboard/openWorkspacesConnecting.test.js"
     ],
     "evidence": [
-      "scripts/run-open-project-safety-checks.js"
+      "scripts/run-open-project-safety-checks.js",
+      "src/openWorkspaces/dashboardController.ts",
+      "src/webview/webviewContent.ts"
     ]
   },
   {
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "131b4c3b22bfaf2e7807a4a1379366db9d4ebc54",
+        "head": "a7613039cd4196cb6f43c022d15eea1483204eaf",
         "ignoredDocumentationCommits": [
             "dd86a325e3db5aa013ffa18a647e67e9fa279c10",
             "0b0e12b4d97ec7d77a8a93076459fdaad2e52070",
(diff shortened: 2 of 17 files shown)


hzcheng:
lgtm

PR #153, opened 2026-08-06
Keep src/dashboard.ts inside the coverage report

hzcheng:
## Summary

`src/dashboard.ts` — 2296 lines, the largest file in the extension — was absent from the coverage report entirely. Every one of the other 74 `src/` modules over 250 lines appears in it; that one did not.

The cause is `tests/contract/aiSessions/runtimeComposition.test.js`. It is the only test that executes `src/dashboard.ts`, it does so by spawning the production activation harness in a subprocess, and it cleared `NODE_V8_COVERAGE` for that child. c8 collects subprocess profiles through exactly that variable, so the file was invisible.

The clearing was copied from `tmuxSmokeHarness.test.js`, which genuinely needs an uninstrumented child — it drives real tmux and asserts `v8Coverage: null` explicitly. Nothing required the same of the activation harness, and no test asserted it. That harness now inherits the variable; the tmux smoke harness is untouched.

## What the file's coverage actually is

| metric | src/dashboard.ts |
|---|---|
| lines | 76.91% (1766/2296) |
| branches | 83.71% (257/307) |
| **functions** | **36.00% (166/461)** |

The shape is informative. Line coverage is respectable because `activate()` and `initializeDashboard` run top to bottom under the harness. Function coverage is not, because 295 functions are never invoked — and 271 of those sit inside the `initializeDashboard` construction section. They are the callback properties of the options objects being wired up (`onDiagnostic`, `onConflict`, `confirmDelete`, `createId`, `renderAiPanel`, …). The wiring executes; the things it wires almost never fire.

Repository totals move because 2296 previously invisible lines re-enter the denominator: functions 86.79% → 80.81%, statements 85.07% → 84.40%, branches 76.93% → 78.10%. All stay above `.ci/coverage-baseline.json`, which is unchanged.

## Guard

A percentage baseline cannot catch this class of regression by itself: dropping a file out of the denominator usually makes the totals look *better*, which is why this went unnoticed since 2026-07-23. `check-coverage-baseline.js` now fails when a file reachable only through a subprocess harness is missing from the report, with a message naming the cause.

Mutation-verified: reverting the harness change makes the guard fail with `src/dashboard.ts is missing from the coverage report; its executing process must inherit NODE_V8_COVERAGE`, and restoring it makes it pass.

`main()` was split so the decision path is reachable from unit tests rather than only from a real repository layout on disk.

## Skill harvest

no skill change — the miss was a gap in repository tooling, not in Agent workflow guidance. The existing skills already require CI-reachable coverage and mutation-sensitive guards; this PR supplies the missing mechanical check rather than a new instruction.

## Verification

- `npm run test:ci:linux` — exit 0
  - deterministic: 615 unit / 719 contract / 336 integration; browser 140
  - changed-line coverage 83.33% (30/36) against `origin/main`
  - coverage baseline passed with the new instrumentation assertion active
- activation harness run 5× uninstrumented and 3× instrumented: 20/20 each time, no timing flakiness from the added instrumentation
- behavior `COVERAGE-BASELINE-INSTRUMENTATION-001` registered in `docs/testing/behavior-contracts.json`

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -3591,6 +3591,20 @@
     ]
   },
   {
+    "id": "COVERAGE-BASELINE-INSTRUMENTATION-001",
+    "domain": "architecture",
+    "title": "Files reachable only through a subprocess harness stay inside the coverage report",
+    "priority": "P1",
+    "status": "automated",
+    "owners": [
+      "tests/unit/tooling/coverageBaseline.test.js"
+    ],
+    "evidence": [
+      "scripts/check-coverage-baseline.js",
+      "tests/contract/aiSessions/runtimeComposition.test.js"
+    ]
+  },
+  {
     "id": "COVERAGE-CHANGED-CODE-001",
     "domain": "architecture",
     "title": "Pull requests keep instrumented changed executable lines at or above the reviewed coverage threshold",
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "cc80a21cd6af2a684b60369cc3195bd3d9bf5f49",
+        "head": "94c139ff153d9c39e809814de35605d62b44e363",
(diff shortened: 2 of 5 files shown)


hzcheng:
lgtm

PR #155, opened 2026-08-07
Improve AI Conversation switching across providers

hzcheng:
## Summary

- reuse one correlated provider snapshot for initial Conversation rendering across Codex, Kimi, and Claude
- retain the existing Webview document while switching sessions and providers, with generation-safe atomic state replacement
- synchronize terminal authority without stealing Conversation focus and roll back when a runtime disappears or changes during switching
- add cross-provider performance, browser, runtime-focus, and race regression coverage

## Verification

- `npm run test:ci:linux`
- `SKIP_NPM_CI=1 npm run install-local` (both VSIX installations verified byte-for-byte)
- real-session snapshot probes: Kimi ~996 ms, Claude ~300 ms, Codex ~1694 ms, each using one shared revision
- final read-only review: no Critical or Important findings

## Skill harvest

- `none` — the review and CI friction was already covered by the repository's existing review/fix workflow guidance; no reusable skill change was justified

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -4116,6 +4116,11 @@
     "owners": [
       "scripts/run-conversation-performance-checks.js",
       "tests/browser/conversationViewer.test.js",
+      "tests/contract/aiSessions/claudeConversationAdapter.test.js",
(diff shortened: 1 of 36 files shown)


hzcheng:
lgtm

PR #157, opened 2026-08-07
fix: allow global active session navigation

hzcheng:
## Summary

- make Previous Active Session and Next Active Session globally discoverable by removing the AI Conversation focus enablement constraint
- switch from the currently open Conversation target even when keyboard focus is in the terminal, Command Palette, or another surface
- keep terminal/card authority synchronized and restore keyboard focus to AI Conversation after switching
- update user feedback, regression ownership, and the main-capability audit

## Skill harvest

no skill change — the user correction changed product behavior, while the existing worktree, RED-first regression, verified installation, and PR workflow skills already covered the execution process.

## Verification

- npm run test-compile
- node --test tests/contract/dashboardBoundaries.test.js
- node --test tests/integration/dashboard/conversationOpenLatest.test.js
- git diff --check origin/main...HEAD
- SKIP_NPM_CI=1 npm run install-local (Agent Pivot 1.0.4 and Attention UI Bridge 1.0.1 installed; packaged-to-installed bytes verified)

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -4588,7 +4588,7 @@
   {
     "id": "CONVERSATION-ACTIVE-SESSION-NAVIGATION-COMMANDS-001",
     "domain": "session",
-    "title": "Previous and Next Active Session commands synchronize the authoritative active terminal and card while restoring keyboard focus to AI Conversation",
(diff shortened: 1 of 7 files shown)


hzcheng:
lgtm

PR #158, opened 2026-08-07
fix: strip inline emphasis tags from Mermaid diagram labels

hzcheng:
## Summary

AI providers frequently emit inline emphasis tags such as `<b>` or `<i>` inside Mermaid diagram labels. The conversation viewer initializes Mermaid with `htmlLabels: false` and `securityLevel: 'strict'`, so those tags were escaped and painted literally into the SVG text.

The viewer now strips inline `b`, `strong`, `i`, and `em` tags from the diagram source before rendering (line-break tags like `<br/>` remain untouched because Mermaid handles them natively). The raw source is still used as the `preserve()` matching key, and the sanitized source feeds the generated alt text.

## Skill harvest

no skill change — fix is a localized source sanitization plus one browser regression test; existing RED/GREEN and audit workflow skills covered the work.

## Verification

- New browser regression test `WEBVIEW-AI-SESSION-CONVERSATION-VIEWER-001 strips inline emphasis tags from Mermaid labels instead of rendering them literally` fails before the fix (mutation-verified by stashing the implementation) and passes after it.
- Full `tests/browser/conversationViewer.test.js` suite: 71/71 pass.
- `npm run test:behavior-contracts`, `npm run lint:ci`, and `git diff --check` pass.
- `npm run test:ci:linux` passes on the final commits.

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "8bef5e6e4ecffde948dd4fbb94e273fff78426fa",
+        "head": "eec04834cd255f0775672fd7f0fa708d70bc041c",
         "ignoredDocumentationCommits": [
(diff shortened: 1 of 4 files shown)


hzcheng:
lgtm

PR #161, opened 2026-08-07
Fix Claude compact lifecycle status

hzcheng:
## Summary

- Treat Claude manual `compact_boundary` records as idle/stopped so compacted sessions do not remain stuck in the running state.
- Keep automatic and unknown compact triggers running, and ignore synthetic `isCompactSummary` user records so record order or timestamp ties cannot overwrite the boundary decision.
- Add regression coverage for manual, automatic, unknown-trigger, resumed-assistant, and out-of-order compact records.

## Validation

- `npm run test:ci:linux`
- Focused lifecycle, provider, and attention suites: 93 passed
- Real Claude transcript validation: manual boundaries stopped; automatic boundaries remained running
- `git diff --check`
- Local VSIX installation and installed-byte verification for Agent Pivot 1.0.4 and Attention UI Bridge 1.0.1
- Final read-only integration review: no Critical, Important, or Minor findings

## Skill harvest

No skill change was justified. The existing provider-adapter skill already requires a real-data shape census, real-layout fixtures, and compiled-adapter validation; the initial sampling gap was a compliance issue rather than a missing reusable rule.

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "18a6618d6dbd214d0383938fac41f1cb3730f430",
+        "head": "687e3f6959ffce62db59daed3fc785a78a284ed1",
         "ignoredDocumentationCommits": [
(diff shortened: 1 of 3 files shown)


hzcheng:
lgtm

PR #162, opened 2026-08-07
feat: collapse completed-turn work behind a Worked-for row

hzcheng:
## Summary

Mirrors the Codex app conversation UX: once a turn's final answer lands, its tool, thinking, and progress entries collapse into a single row labeled **Worked for 1m 20s** (plain **Worked** when the provider exposes no timing, e.g. Codex). The row heads the work group accordion-style, so expanding reveals the entries below the toggle without ever moving it under the pointer, and the row sits flush with the message column (expanded entries keep their indent as children of the header); the expanded state survives live refreshes within the session generation. Turns still in progress stay expanded, matching the Codex `hasFinalAssistantStarted` collapse trigger.

- Claude and Kimi adapters stamp each interaction's `completedAt` from the last contributing provider event; page interaction states carry the turn timing so the viewer renders the divider between the work entries and the first assistant answer.
- The row is a normal `data-message-id` article (`<interactionId>:worklog`), so it rides the existing reconcile/signature preservation path.
- Webview owns the transient expanded state (Map keyed by interaction id, reset on session generation), applies `hidden` to pre-answer work entries, and sets `aria-expanded` after sanitizing; the sanitizer allowlist gains `button`.
- Codex turns show the fallback label because the app-server exposes no turn timing yet.

Behavior catalog: new `CONVERSATION-WORKLOG-COLLAPSE-001` (P1) with unit, contract, integration, and browser owners.

## Skill harvest

no skill change — the RED/GREEN, webview mutation, and audit workflow skills covered the work; the one execution slip (an over-broad test-file rewrite) was caught by the existing full-file suite re-run and needed no skill iteration.

## Verification

- RED first: all six focused owners failed for the missing behavior before implementation (missing formatter, missing `completedAt`, no row, sanitized button).
- Focused GREEN: unit text/model, Claude/Kimi adapter contracts, integration viewer, browser collapse/toggle/refresh-preservation and live in-progress→complete journey.
- Full touched suites: `conversationText`, `conversationModel`, Claude/Kimi/Codex adapter contracts, coordinator, integration viewer + openLatest, and the 73-test browser viewer file all pass.
- `npm run test:behavior-contracts`, `npm run lint:ci`, `git diff --check` pass; media webview copies byte-identical with `src/webview`.
- `npm run test:ci:linux` passes on the final commits.
- Screenshot-reviewed collapsed and expanded states in the real webview document.


hzcheng:
lgtm

PR #166, opened 2026-08-07
Fix AI Conversation working indicator lifecycle

hzcheng:
## Summary

- Keep readable provider lifecycle signals authoritative during quiet work so running sessions do not age into a false stopped state.
- Follow newly appended running turns without overriding explicit historical navigation, including coalesced refreshes and stale-revision recovery.
- Show Working when the latest in-progress interaction is already rendered even if the selected-input metadata is transiently stale.
- Handle a temporarily unavailable lifecycle signal without throwing, and keep safety fixtures aligned with failure and recovery semantics.

## Skill harvest

no skill change — Existing regression and provider-adapter skills already require rendered-surface coverage, real provider data, and the full branch-level CI equivalent before push; the initial CI failure came from not running that existing gate, not from an instruction gap.

## Verification

- `npm run test:ci:linux`
- `npm run test:safety:run`
- Focused execution-monitor contract RED before the null guard, then GREEN
- `npm run test:behavior-contracts` (41 catalog/audit plus 4 release-journey tests passed)
- Changed-line coverage: 100% (65/65) against `origin/main`
- `git diff --check`
- Final independent read-only review: no Critical, Important, or Minor findings

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -4348,11 +4348,13 @@
   {
     "id": "CONVERSATION-WORKING-INDICATOR-001",
     "domain": "webview",
-    "title": "AI Conversation shows a reduced-motion-safe Working indicator only while the latest response remains in progress",
+    "title": "AI Conversation keeps a reduced-motion-safe Working indicator visible through quiet provider work and newly appended running inputs",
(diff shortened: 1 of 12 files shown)


hzcheng:
lgtm

PR #167, opened 2026-08-07
Fix Codex worklog duration display

hzcheng:
## Summary

- preserve Codex app-server turn timing through the conversation adapter so completed worklogs show their duration
- sum fully timed subagent turns while excluding idle gaps and fail closed to plain Worked when any contributing turn is incomplete
- add real adapter-to-viewer coverage for precise durations, fallbacks, multi-turn dispatches, and malformed timing

## Testing

- npm run test:ci:linux
- npm run test-compile
- node --test tests/contract/aiSessions/codexConversationAdapter.test.js tests/integration/dashboard/conversationViewer.test.js
- git diff --check origin/main..HEAD
- SKIP_NPM_CI=1 npm run install-local

## Skill harvest

No skill change was justified. The task friction around build ordering, review fixes, capability auditing, and installed-byte verification is already covered by the existing repository skills.

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -4050,6 +4050,7 @@
       "tests/unit/aiSessions/conversationText.test.js",
       "tests/unit/aiSessions/conversationModel.test.js",
       "tests/contract/aiSessions/claudeConversationAdapter.test.js",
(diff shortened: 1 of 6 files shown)


hzcheng:
lgtm

PR #169, opened 2026-08-07
Fix duplicate external link navigation

hzcheng:
## Summary

- let VS Code Webview handle HTTPS links through its native navigation path
- keep absolute local file links routed through the Extension Host for exact file and line opening
- add rendered browser coverage that rejects simultaneous native and Host navigation for both initial and sanitized live content

## Skill harvest

no skill change — the only workflow retry was a missing compiled test artifact, and the existing testing guidance already requires compilation before browser tests

## Verification

- `npm run test:ci:linux` (149 Chromium tests; changed-line coverage passed)
- focused rendered HTTPS and local-file link browser tests (3 passed)
- `node scripts/run-dashboard-webview-checks.js`
- `git diff --check`
- `SKIP_NPM_CI=1 npm run install-local` (Agent Pivot 1.0.4 and Attention UI Bridge 1.0.1 installed; packaged and installed bytes verified)

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "6218c88baa1aacc6ad07f611fa38bf52ef999ab2",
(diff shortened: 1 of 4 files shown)


hzcheng:
lgtm

PR #170, opened 2026-08-07
feat: add in-conversation find bar with match highlighting

hzcheng:
## Summary

- Adds a Ctrl/Cmd+F find bar to the AI conversation viewer: a compact overlay anchored to the top-right of the conversation viewport with a search field, live match count, and previous/next/close controls
- The query matches case-insensitively across the visible conversation text (collapsed worklog content is skipped), including phrases that span inline-code boundaries; runaway result sets stop at an announced 999+ cap and queries without matches flag the field red
- Every match paints through CSS custom highlights (`conversation-find` / `conversation-find-current`) with a distinct current match that scrolls into view; Enter / Shift+Enter and the bar controls step with wrap-around, and the count announces the position through a polite status role
- The query reapplies after every authoritative page refresh, keeping the current position clamped against the new match set without yanking the scroll position
- Escape closes the bar, restores focus, and clears all highlights; reopening keeps and selects the previous query, and a text selection inside the conversation seeds the search

## Behavior contracts

- Adds `CONVERSATION-FIND-001` (P1, automated): integration coverage for the find bar markup and the script-order chain, browser coverage for Ctrl+F open / Escape close, case-insensitive and cross-node matching, highlight registration, Enter/Shift+Enter/button wrap-around, scroll-into-view, no-results flagging, the 999+ cap, refresh reapplication, and top-right overlay geometry

## Verification

- `npm run test:ci:linux` passes, including release packaging (the new script ships in the VSIX) and changed-line coverage 100% (33/33)
- `tests/browser/conversationViewer.test.js` 157/157 with six new find tests, `tests/integration/dashboard/conversationViewer.test.js` 385/385
- Rendered-output screenshots reviewed at 700px and 320px panel widths: the bar hugs the viewport's top-right, stays usable at minimum width, and shows no clipped or truncated labels
- `git diff --check` clean; `media/conversationFindScripts.js` is byte-identical to `src/webview/conversationFindScripts.js`

## Skill harvest

- Updated `.skills/resilient-webview-mutation-protocols` with the new-Webview-script registration checklist (document-builder script tag, `.vscodeignore`, release-packaging entries, browser harness route, contract evidence) derived from this branch's six registration touch-points; validated with `tests/unit/tooling/repositorySkills.test.js`


hzcheng:
lgtm

PR #173, opened 2026-08-08
feat(conversation): Workspace-scoped notes in the Comments panel, with a full comment-stack consolidation

hzcheng:
## Summary

**Feature: project-scoped Workspace notes in the AI Conversation Comments panel.** The panel now opens with a Workspace section above the Session list: quick-capture notes with free-form tags, a panel-wide filter bar (status + tags), collapsible groups with live open/total counts, a draggable sash between the two independently scrolling regions, and header-embedded actions (compose, send, clear done, clear all). Workspace notes are shared by every Session of the project (persisted under `project-comments/v1` keyed by projectId), keep a dispatch history, and stay open until deliberately marked done; selected conversation text can be saved straight into a note as a source snapshot. Session comment cards gain the same unified tag/status chip row and drag-to-reorder.

**Refactor: full review and consolidation of the comment feature** (both stacks), fixing 14 confirmed bugs and removing the parallel-implementation duplication:

- `commentPrimitives.ts` — one `CommentError` base class (both public error classes extend it, so `instanceof` works across stacks) plus single copies of the validation/text/tag/reorder/clear primitives (previously up to 4 copies).
- `snapshotFileStore.ts` — generic atomic snapshot persistence (digest-addressed files, temp-write + rename, fail-soft reads); both stores became thin wrappers, rebind copy stays session-only.
- `queuedCommentController.ts` — the shared resilience protocol (serialized queue, optimistic revisions, settlement replay with LRU idempotency, persist → commit → rollback, rebuild-on-undeliverable). Both controllers lost ~75–80% of their lines to it.
- Webview: `sessionStack`/`projectStack` config objects collapse ~20 near-duplicate function pairs (validators, pending gate, posting, settlement, tag rows/chips, icon buttons, drag & drop); the script shed its dead code and shrank by ~220 lines.
- Protocol: operation lists are single-sourced from model constant arrays (types derive via `typeof`, runtime checks reuse the same arrays), and the order-sensitive terminal negative guard is now an explicit positive branch.
- Notable fixes: pending gate missed `<input>` controls, project re-render dropped the disabled state mid-flight, `resetSession` leaked the tag editor, the shared status line raced between stacks (Workspace results are now prefixed), `clearAll` could be blocked by a corrupt in-memory entry, locate raced queued mutations (now queued), and cross-class submission errors degraded to `failed` (now preserved via the shared base).
[...]

PR #174, opened 2026-08-08
refactor(conversation): consolidate the bookmark stack onto the shared mutation pipeline

hzcheng:
## Summary

Follow-up to #173: the bookmark stack was the third instance of the same queue/settlement/persist pattern and is now migrated onto the shared modules built there. Pure internal consolidation — no user-visible change.

- `snapshotFileStore.ts` is generalized on the payload shape (`payloadKey`/`itemsOf`/`buildSnapshot`/`maxSnapshotBytes` hooks), and `bookmarkStore.ts` becomes a thin wrapper (306 → ~110 lines). The on-disk format (`conversation-bookmarks/v1`, `interactionIds` field) is unchanged — zero migration.
- New `queuedMutationController.ts` extracts the mid-level pipeline shared by all three viewer-state controllers: serialized operation queue, optimistic revisions, idempotent settlements with replay, persist/commit gate helpers, and the rebuild-on-undeliverable fallback. `QueuedCommentController` was re-based onto it with behavior preserved point-for-point, and `bookmarkController.ts` was migrated (274 → ~160 lines).
- Three latent edge inconsistencies aligned to the comment stacks' semantics, each pinned by a unit test: settlement keys no longer include the subscription generation (a cross-generation retry now replays the true outcome instead of `stale`; the webview drops stale-generation messages on its own), redundant bookmark toggles settle successfully without rewriting identical bytes to disk, and replayed settlements no longer re-trigger a document rebuild on delivery failure (the original publish already attempted it).
- First dedicated unit suite for the bookmark controller (9 tests: toggle flow, stale defense, replay, budget limit, freeze/reset, restore gating, no-op persistence skip, IO failure mapping), registered as an owner of `CONVERSATION-OUTLINE-BOOKMARKS-001`.
- Public API surface unchanged: viewer.ts, composition.ts, dashboard.ts, and sessionRebindCoordinator.ts required no edits.

## Skill harvest

no skill change — this cycle's friction (an `Array.every(fn)` arity trap caught immediately by the store tests, and a JSON contract-file reformat that was reverted in favor of surgical edits) was reviewed against the decision standard: both are one-off items already guarded by existing checks. Recorded as `Skill-Harvest: none` in the capability audit commit (8b1ff5e).

## Verification

- `npm run test-compile`; unit **674/674** (665 + 9 new); integration **385/385**; browser **162/162**; `lint:ci`; `test:behavior-contracts`; `brand:check`; coverage baselines (`test:coverage:ci`, changed-line coverage 96.71%).
- Capability audit regenerated and green (`8b1ff5e`); all five implementation commits assigned to `MAIN-AI-SESSION-CONVERSATION-OUTLINE`.
- Packaged and installed locally (`SKIP_NPM_CI=1 npm run install-local`, byte-verified) and accepted through a manual regression pass (bookmarks, both comment stacks).

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -4614,11 +4614,13 @@
     "priority": "P0",
     "status": "automated",
     "owners": [
+      "tests/unit/aiSessions/conversationBookmarkController.test.js",
       "tests/unit/aiSessions/conversationBookmarkStore.test.js",
       "tests/integration/dashboard/conversationViewer.test.js",
       "tests/browser/conversationViewer.test.js"
     ],
     "evidence": [
+      "src/aiSessions/conversation/bookmarkController.ts",
       "src/aiSessions/conversation/bookmarkStore.ts",
       "src/aiSessions/conversation/viewer.ts",
       "src/webview/conversationOutlineScripts.js",
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "84daa1256b925387cc57a8d3af4c04c016ad3c5c",
+        "head": "fe9935f1097409d38cb6f6bad191528e83aa4279",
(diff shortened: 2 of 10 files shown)


hzcheng:
lgtm

PR #180, opened 2026-08-08
fix: resolve relative Kimi Shell cd targets for telemetry worktree branches

hzcheng:
## Summary

A Kimi session operating in a worktree via `cd .worktree/...` showed the launch repo's branch (`main`) in the conversation telemetry forever. Root cause: the Shell `cd` mining only kept **absolute** targets, silently dropping every relative one. Verified against the installed kimi-cli source that the Shell tool is stateless per command (`cwd = session.work_dir`), so relative targets now resolve against the session working directory, chaining within a single command; `~` and `$VAR` forms stay unresolvable by design.

Also verified the other providers do **not** share the bug: Claude's per-record `cwd` follows the persistent shell (verified end-to-end on a real worktree session, resolves the worktree branch), and Codex reads explicit absolute `workdir` fields from exec records.

## Test plan

- [x] RED-first contract test (relative cd dropped → `main`), then GREEN
- [x] Newest-resolvable-wins fallthrough pinned (deleted newest worktree → next mined path)
- [x] Compiled adapter + real git resolver against the reported session: now resolves `fix/marketplace-compliance`
- [x] unit 686 / contract 783 / integration 387 / browser 170
- [x] `lint:ci`, `brand:check`, behavior contracts + capability audit
- [x] local install byte-verified

Contract: `CONVERSATION-TELEMETRY-001`.

## Skill harvest

Updated `.skills/developing-provider-conversation-adapters`: the Kimi Shell tool is stateless per command (cwd = session work_dir, source-verified) — durable provider knowledge for any future cwd-dependent parsing.

### DIFF
--- a/.skills/developing-provider-conversation-adapters/SKILL.md
+++ b/.skills/developing-provider-conversation-adapters/SKILL.md
@@ -37,7 +37,7 @@ Green self-authored fixtures alone are not evidence.
 
 | Provider | Session source | Subagents |
 |---|---|---|
-| Kimi | `<kimiHome>/sessions/<workdirHash>/<sessionUuid>/wire.jsonl` | `<sessionDir>/subagents/<id>/{meta.json,wire.jsonl}`; meta carries explicit `status` + `created_at` |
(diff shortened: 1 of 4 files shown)


hzcheng:
lgtm

PR #181, opened 2026-08-08
fix: marketplace republication compliance hardening

hzcheng:
## Summary

Deep publication-safety remediation after the Marketplace publisher suspension for identity/metadata confusion. Every fix ships with an automated guard so CI and `vscode:prepublish` reject regressions:

- **VSIX packaging leak**: `.vscodeignore` now excludes `.worktree/**` — a local release packaged from the primary checkout would otherwise embed 20k+ internal files (nested worktrees, process docs, tests, nested `node_modules`). Also excludes `artifacts/**`, `*.vsix`, and `.gitattributes`; release packaging checks assert the rules and forbid those prefixes inside the real archives.
- **Residual "Dashboard" self-references**: prompt webview announcements, the tablist aria label, the diagnostics log prefix, and one README bullet now say Agent Pivot; `the Dashboard` / `Dashboard views` / `[Dashboard]` join the brand-identity stale tokens.
- **Changelog hygiene**: the unpublished pre-release history moved from `CHANGELOG.md` to `docs/development-history.md` (archived bytes sha256-locked). The published changelog no longer shows version inversions or a previous identity, and the brand scanner covers the whole file again.
- **UI Bridge metadata**: added `homepage`/`bugs`; README links are absolute GitHub URLs; packaged READMEs are guarded against parent-relative links that break on the Marketplace.
- **Webview CSP**: dashboard `img-src` is scoped to the webview origin, `data:`, and `https:` (no `*`); script `'unsafe-inline'` is replaced by per-render nonces in both the dashboard and boot documents; the inline stylesheet `onload` handler is now a nonce'd listener with sheet/error fallbacks.
- **Hygiene**: dropped `DemoProjects.json` (upstream author's demo paths) and a hardcoded developer-machine path in a spike.

Deferred to follow-up PRs (assessed, out of scope here): build-artifact and contribution-ID renames (`dist/dashboard.js`, `webviewDashboard*.js`, view id `agentPivot.dashboard`), and the Sharingan animation assets (third-party IP evaluation).

## Skill harvest

updated .skills/review-fix-commit-loop — moving the unpublished history broke a changelog-anchored check that only surfaced when its own suite ran; the skill now requires grepping `scripts/` and `tests/` for anchored readers before moving or deleting repository content. The same decision is recorded as the Skill-Harvest trailer of the capability audit commits.

## Verification

- `npm run test:ci:linux` — green on the rebased branch (brand checks, behavior contracts, coverage baselines, browser/safety/dashboard suites, release notes, release packaging).
- `npx @vscode/vsce ls` with seeded `.worktree/` and `artifacts/` leak files — 82 packaged entries, zero leaks.
- Release packaging checks against the real VSIX archives — passed.
- Final read-only integration review of the complete diff — no Critical/Important findings left open.


hzcheng:
lgtm

PR #182, opened 2026-08-08
fix(conversation): keep the Workspace composer fully visible in a full group

hzcheng:
## Summary

Fix the Workspace note composer collapsing into a clipped sliver once the Comments side panel's Workspace group fills up with cards.

- Root cause: `.conversation-project-comment-composer` had no flex declaration, so it defaulted to `flex: 0 1 auto`; combined with `overflow: hidden` its automatic minimum size is zero, letting it absorb all negative free space when the never-shrinking card list (`flex: 0 0 auto`) fills the bounded group region. Only the textarea row stayed visible — the label and the Cancel/Add actions row were clipped away.
- Fix: give the composer `flex: 0 0 auto`, the same protection the Session composer already had (`media/conversationViewer.scss`), and recompile `media/conversationViewer.css`. The overflowing card list now scrolls inside the group region underneath a fully rendered composer.
- Test: new `PROJECT-COMMENTS-UI-001` browser test seeds a 16-card group, reopens the composer, and asserts the rendered geometry (group overflows, composer content unclipped, actions row contained, composer inside the group region and clear of the first card). It fails on the unfixed styles with `contentUnclipped: false` / `actionsContained: false`.

## Skill harvest

no skill change — the regression flowed through the existing fixing-regressions-with-ci and review-fix-commit-loop checklists (RED first, SCSS recompile, rendered-pixel screenshot) with no friction or corrections.

## Verification

- RED: new test fails on unfixed styles with the exact reported symptom (composer clipped, actions row cut off)
- `node --test tests/browser/conversationViewer.test.js` — 94/94 pass (rebased on `d1b619d`)
- `tests/integration/dashboard/styles.test.js` — 12/12 (SCSS ↔ compiled CSS consistency)
- `npm run test:behavior-contracts`, `npm run lint`, `git diff --check` — clean
- `npm run test:ci:linux` — full branch suite green (exit 0), changed-line coverage 100%
- Rendered-pixel screenshots at 700px and 480px viewports: full composer (label, textarea, actions row) visible above the scrolling card list
- Manually installed the packaged VSIX (`agent-pivot-1.0.4.vsix`, hash-verified) and confirmed the fix in the live Comments panel


hzcheng:
lgtm

PR #183, opened 2026-08-08
feat: replace bundled third-party artwork with user-supplied custom running images

hzcheng:
## Summary

The bundled anime artwork running-animation options were third-party intellectual property (the SVGs are fan recreations of protected designs, so the upstream CC BY-SA grant cannot cover the underlying characters) — a publication and takedown risk for the Marketplace account. This PR removes the bundled artwork while keeping the feature alive as a user-supplied capability:

- **New settings** `agentPivot.aiSessionRunningCardCustomImage` and `agentPivot.aiSessionRunningIconCustomImage` plus a `custom` choice in both running-animation enums: users point at a local image file (SVG, PNG, GIF, WebP, or JPEG up to 256 KB, absolute or `~` path). The file is read locally and never uploaded.
- **Resolver** `src/webview/runningAnimationImages.ts`: path expansion, file/type/size/readability guards, mtime-keyed caching; unresolvable `custom` selections and stale removed enum values fall back to the default `current` animation, so existing user settings keep working.
- **Rendering**: images are injected as document-level CSS variables (`--agent-pivot-running-card-image` / `--agent-pivot-running-icon-image`) consumed by the existing `data-session-fx` / `data-session-icon-fx` rules; full and incremental render paths both resolve effective values through one helper.
- **Removed**: `media/sharingan/`, the retired enum options, the attribution section in THIRD_PARTY_NOTICES.md, and the asset-integrity test; README now documents the custom-image slots. The brand identity scanner rejects the retired artwork names so they cannot be reintroduced.

## Skill harvest

no skill change — the applicable pitfalls (rebuild webview assets with `npx gulp --production`, grep compiled CSS for new selectors, chain verification before commits) were already recorded in the existing skills and were followed this round.

## Verification

- `npm run test:ci:linux` — green on the rebased branch (includes behavior contracts, coverage baselines, browser/safety/dashboard suites, release packaging).
- New `CUSTOM-RUNNING-IMAGE-RESOLVER-001` unit tests cover path/type/size guards, tilde expansion, mtime caching, and fallback behavior; `CUSTOM-RUNNING-IMAGE-001` integration tests assert CSS-variable injection and `current` fallback in the full render.
- `npx @vscode/vsce` packaging checks confirm the retired assets are gone from the VSIX and the expected entry list stays exact.


hzcheng:
lgtm

PR #184, opened 2026-08-08
feat: cycle running AI sessions across windows with Next Running Session

hzcheng:
## Summary

Adds a new `Agent Pivot: Next Running Session` command that cycles AI sessions whose execution state is `running` (a turn is actively processing), one jump per invocation:

- **Local jumps** focus the session's terminal/tmux window and open its conversation viewer, riding the same wiring as the attention queue jump. Locals sort by an immutable `provider:sessionId` key and lead the queue, matching the attention queue's local-first order.
- **Cross-window jumps** use a new one-shot focus mailbox on the Open Windows bridge channel: the origin window writes a focus request (`open-workspaces/running-focus/v1`), the winning window's bridge delivers it to its main extension, which lands directly on its next local running session — a single keypress lands on a working session in the other window. When the bridge predates the mailbox channel, the request command is missing and the jump degrades gracefully to a plain window switch.
- **Resilient cursor**: the cycle advances even when a jump fails, so a session that ends mid-cycle is skipped with a warning instead of trapping the user; a vanished cursor restarts at the queue head.
- No default keybinding (palette-first, like Previous/Next Active Session).

## Implementation

- `src/aiSessions/runningQueue.ts` — pure queue builder + cursor picker.
- `src/dashboard/runningSessionJump.ts` — jump handler closure shared by the command (`all` scope) and the mailbox receiver (`local` scope), one shared cursor.
- `src/openWorkspaces/runningFocusProtocol.ts` — strictly validated request/outcome shapes, 60s lease, bounded mailbox.
- Bridge extension: `openWorkspaceRunningFocusStore` (atomic per-request files, lease sweep, pending cap) + `openWorkspaceRunningFocusCoordinator` (winner-only delivery via the same priority order as direct navigation, consume-on-delivery, retry-until-lease) + `OpenWorkspaceCoordinator.isNavigationWinner`.
- Main extension: `OpenWorkspaceBridgeClient.requestRunningFocus` (correlated outcome, `false` on any failure for graceful degradation) and a deduplicated `runningFocusRequested` receiver; pre-adoption requests are logged and dropped in `EarlyOpenWorkspaceBridge`.
- No protocol version bump and no capability-flag changes: the mailbox lives in its own directory and detection is command-presence based, so mixed extension generations keep working (degraded hop).

## Verification

- Full `test:ci:linux` gate chain green after rebasing onto the latest `main` (3b7402e): unit/contract/integration under coverage, browser (172), safety, dashboard webview checks, architecture guards, release notes, release packaging; changed-line coverage 92.92%.
- New contract tests: queue/cursor/jump semantics (`runningSessionJump.test.js`), protocol/store/coordinator/winner resolution (`runningFocus.test.js`), client send/receive/degradation (`bridgeClient.test.js`), pre-adoption drop (`earlyBridge.test.js`), manifest contribution without keybinding (`dashboardBoundaries.test.js`).
- Behavior catalog: `AI-SESSION-NEXT-RUNNING-COMMAND-001` plus `OPEN-WORKSPACE-RUNNING-FOCUS-{PROTOCOL,STORE,COORDINATOR,CLIENT}-001`.
- Real-machine acceptance passed with both VSIX builds installed (main extension into the vscode-server, bridge on the local UI host): local cycling, single-press cross-window landing, and empty-queue messaging verified across reloaded windows.

## Skill harvest

Updated `.skills/registering-extension-commands` (new): this round failed two late gates because the command surface was synced to the four remembered registries but not to `scripts/run-dashboard-webview-checks.js` or the reviewed VSIX list in `scripts/run-release-packaging-checks.js`. The skill replaces memory with an executable sibling probe (`grep -rln 'agentPivot.nextAttentionSession' src/ tests/ scripts/ package.json`) and covers the packaged-`out/` module and bridge `tsconfig.json` include syncs, with owner coverage in `tests/unit/tooling/repositorySkills.test.js`. Two other frictions (piping gate output through `tail`, packaging before the latest-base rebase) were already covered by `review-fix-commit-loop` and `publishing-and-merging-github-prs` — compliance gaps, not instruction gaps, so no text changes there.


hzcheng:
lgtm

PR #188, opened 2026-08-09
feat: add AI session quick switching and harden conversation navigation

hzcheng:
## Summary

- add a global AI session picker and an MRU toggle for local sessions, including cross-window handoff for running sessions
- sample Direct and tmux focus independently of attention state so active sessions feed the MRU reliably
- confirm empty speculative conversation snapshots before reporting an active Kimi session as conversation-less
- refresh Codex worktree telemetry from the latest rollout working directory while retaining cached model and quota telemetry

## Testing

- `npm run test:ci:linux`
- focused AI session quick-switch, conversation composition, and Codex telemetry tests
- real Codex rollout probe verified the feature worktree and branch independently from the launch checkout

## Skill harvest

No skill changes were justified. The existing regression, provider-adapter, worktree, review, and publishing skills covered the required RED-before-fix checks, real-provider verification, capability audit, and publication sequence.

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -797,6 +797,24 @@
     ]
   },
   {
+    "id": "AI-SESSION-QUICK-SWITCH-COMMANDS-001",
+    "domain": "session",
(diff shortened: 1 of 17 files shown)


hzcheng:
lgtm

PR #189, opened 2026-08-09
feat: make attention clearing on session jumps optional

hzcheng:
## Summary

- add `agentPivot.aiSessionAttention.clearOnNextSession`, disabled by default
- keep attention indicators unread after Next Attention Session unless the setting explicitly opts in, while preserving card-click acknowledgement
- decouple the navigation cursor from acknowledgement so repeated jumps continue cycling through unread sessions
- advance past unavailable sessions instead of trapping the cursor on a stale target

## Testing

- `npm run test:ci:linux`
- focused attention status-bar and dashboard boundary contract tests
- local VSIX package and install verification in the active Dev Container
- changed-line coverage: 100% (24/24)

## Skill harvest

No skill changes were justified. The only additional finding was a product behavior requirement—navigation must keep advancing when acknowledgement is disabled—and it is now captured by the attention queue contract tests; the existing worktree, Webview mutation, review, installation, and publishing workflows were sufficient.

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -755,7 +755,7 @@
   {
     "id": "ATTENTION-STATUS-BAR-QUEUE-001",
     "domain": "attention",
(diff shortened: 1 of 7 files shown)


hzcheng:
lgtm

PR #192, opened 2026-08-09
fix: float the sidebar dashboard header on a single separator line

hzcheng:
## Summary

User-reported visual awkwardness at the top of the sidebar: the header read as two edge-hugging bars stacked under the view splitter. Measured on the real rendered document:

- the settings button overrode its sibling 4px gap down to 0px and sat flush against the collapse-all button;
- the tab strip padded its cells 10px deeper than the search row, so the two rows shared no edge;
- the tab strip border and the sticky header shadow drew two hairlines 8px apart with a dead band between them, and the active-tab underline anchored to the upper one;
- header content started only 8px below the splitter and kissed the panel edges.

Fix (sidebar mode only; editor-mode layout unchanged apart from the uniform button gap):

- every header button keeps the same 4px gap (dropped both `margin-left` overrides);
- header content floats 12px below the splitter and 6px off each panel edge, so the search row and tab strip share symmetric insets;
- the header shadow is the single full-width separator: the tab strip's own border and highlight hairline are dropped, and the active-tab underline lands exactly on the remaining line.

The final spacing direction ("floating" content with a full-width separator) was chosen by the user from four rendered candidate designs.

## Verification

- RED: new `WEBVIEW-SIDEBAR-HEADER-LAYOUT-001` browser test failed on the old CSS for the reported reasons (0px settings gap vs 4px sibling gap; tab strip inset past the search row edge; 8px double-separator band; detached underline; 0px header inset). Mutation sensitivity was exercised during iteration by re-measuring every candidate against these assertions.
- GREEN: focused test passes at 315px and 240px sidebar widths; full browser suite 175/175; `test:behavior-contracts`, `lint:ci`, dashboard webview checks all pass.
- `npm run test:ci:linux` green end to end (including release packaging).
- Packaged and installed locally; installed `media/styles.css` byte-verified against the VSIX and confirmed to contain the new rules; user reviewed the result in a real window and approved the design.

## Skill harvest

none — the round trip was requirements clarification (which visual element felt off), not a workflow-rule gap; no existing skill was wrong or missing a verifiable rule.


hzcheng:
lgtm

PR #193, opened 2026-08-09
fix: handle real-world skill store structures (block scalars, sub-skills, compact folders)

hzcheng:
## Summary

Verified against a real 173-skill store (vendored repos cloned wholesale into `~/.skills`, deep nesting, bundled sub-skills), four structural gaps surfaced and are fixed:

- **frontmatter**: YAML block scalars (`description: >-` folded, `|` literal) are parsed instead of surfacing the indicator as the description; a UTF-8 BOM before the fence is tolerated. Vendored Google skills previously showed a raw `>-` in every row.
- **discovery**: the central-store walk now descends past a skill directory into subtrees that hold more skills (`dms-assistant/mysql` and friends were invisible — 169 → 173 records), without registering helper dirs (`scripts/`, `references/`) as folder nodes.
- **folder tree**: single-child chains compact into one row targeting the deepest real directory (`google/skills/skills/cloud` instead of four nested rows), and empty subtrees deeper than one level stay out of the tree unless their parent directly holds skills, so vendored repo internals (`plugins/`, `docs/`, `scripts/`) no longer pollute the panel.
- **sticky folder headers** pin only the top two *visual* levels (post-compaction), so deep trees no longer stack a chaotic pinned staircase.

Also documents the SKILLS panel rework + these fixes in the changelog and README.

## Skill harvest

no skill change — the fixes followed the existing worktree/verification/audit flow without friction worth recording; recorded as the `Skill-Harvest: none` trailer of the capability audit commit.

## Verification

- Real-store scan: 173/173 SKILL.md discovered; folded descriptions parse correctly; sub-skills found; vendored empty chains pruned — confirmed in rendered screenshots.
- `node scripts/run-skill-management-checks.js` (incl. new nested-skill and block-scalar fixtures), `node scripts/run-dashboard-webview-checks.js`, `npm run test:release-notes`, `npm run lint`, `git diff --check` — green.
- `tests/browser/**` — 176 passed (incl. new SKILLS-SPLIT-007 sticky-level cap); `tests/contract/**` — 827 passed (incl. new compaction/pruning render contracts).
- Changed-line coverage gate: 97.16% (111/111 previously for the rework) against origin/main — `check-coverage-baseline` + `check-changed-coverage` green.
- Built and installed locally (`agent-pivot-1.1.0.vsix`), installed bundle hash-verified against the build.


hzcheng:
lgtm

PR #194, opened 2026-08-09
Fix cross-window next-session navigation

hzcheng:
## Summary

- Make Next Running Session rotate deterministically across windows.
- Make Next Attention Session continue from the watched session and open the exact requested conversation.
- Serialize cross-window focus handoffs so stale Running requests cannot override newer Attention navigation.
- Refresh the sidebar focus projection after VS Code finishes activating the target terminal, so the focused card matches the tmux/window switch.
- Add regression coverage for repeated and interleaved Running/Attention navigation, including the final focused-card state.

## Testing

- `npm run test:ci:linux` (855 deterministic, 394 integration, and 176 browser tests passed)
- Focused next-session and active-session focus behavior-contract tests
- Release packaging checks
- Local installation and byte verification for both VSIX extensions

## Skill harvest

No skill change. Existing regression guidance already requires checking the final rendered surface; this PR adds the missing CI-owned focused-card assertion.


hzcheng:
lgtm

PR #195, opened 2026-08-09
Centralize session navigation transactions

hzcheng:
## Summary

- Serialize Next Running Session and Next Attention Session through one shared navigation coordinator.
- Route both commands through one local focus executor that keeps terminal focus and conversation open on the same workspace snapshot.
- Preserve focus-phase cursor and MRU ordering when conversation opening fails.
- Add an interleaved Running/Attention regression journey and an architecture guard that prevents the commands from regaining private navigation ownership.

## Scope

This is the first architecture-refactor stage. Queue selection policies, cross-window protocols, persisted formats, provider adapters, and tmux backend behavior remain unchanged. Follow-up stages can consolidate the versioned remote transport after this boundary is proven stable.

## Testing

- `npm run test:ci:linux`
- 861 deterministic tests passed
- 394 integration tests passed
- 176 browser tests passed
- Release packaging passed
- Changed-line coverage: 94.04% (205/218)

## Skill harvest

No skill change. Existing architecture-refactor and review-loop guidance already requires safety-net-first execution, call-site review, partial-failure checks, executable boundaries, and fresh full-suite verification.

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -755,7 +755,7 @@
   {
     "id": "ATTENTION-STATUS-BAR-QUEUE-001",
     "domain": "attention",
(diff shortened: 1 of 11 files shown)


hzcheng:
lgtm

PR #198, opened 2026-08-09
fix: restore Codex live conversation telemetry

hzcheng:
## Summary

- replace `thread/resume` telemetry reads with read-only `thread/read` so active Codex writers no longer suppress branch and context data
- read the current model, context-window usage, and latest quoted workdir from the bounded Codex rollout tail
- add regression coverage for active-writer sessions and document the guarded Codex adapter import boundary

## Skill harvest

updated .skills/developing-provider-conversation-adapters — the architecture guard caught that type-only relative imports also enter the Codex adapter graph, so rollout readers must be injected through composition and checked with the focused architecture gate.

## Verification

- `npm run test:ci:linux`
- real active-writer compiled adapter probe against the current Codex rollout
- `SKIP_NPM_CI=1 npm run install-local` with VSIX-to-installed byte verification
- manual telemetry verification pending after local installation


hzcheng:
lgtm

PR #199, opened 2026-08-09
Serialize every AI session switch command

hzcheng:
## Summary

- route Switch to AI Session and Toggle Last AI Session through the shared session navigation coordinator and single-snapshot focus executor
- queue local and remote Quick Switch actions so an older slow navigation cannot overwrite a newer command
- resolve Toggle Last's MRU target only when its queued transaction starts
- join Previous and Next Active Session command invocations to the same global ordering boundary while preserving the Conversation module's newest-intent, rollback, and keyboard-focus behavior
- expand the architecture guard and controlled mutations to cover every terminal-moving AI Session command

## Regression evidence

Before the fix, a controlled journey started a slow Next Running Session navigation and then completed Switch to AI Session. When the older navigation resumed, it replaced both the terminal and Conversation selected by the newer command. The new contract test failed with settlement order `newer, older`; after the fix it passes with `older, newer`, and the final terminal and Conversation both belong to the newer command.

## Scope

Open Current AI Session Conversation remains conversation-only, and Switch to Open Window remains window-only. Running/Attention queue selection, MRU policy, focus protocols, mailbox formats, attention acknowledgement, tmux behavior, and serialized data are unchanged.

## Verification

- `npm run test-compile`
- 113 focused navigation and Conversation tests passed
- 66 architecture mutation tests passed
- `npm run test:architecture-guards`
- `npm run lint`
- `npm run test:ci:linux`
- changed-line coverage: 100.00% (50/50)
- release VSIX packaging checks passed
- `git diff --check`

## Skill harvest

No skill change was justified. The earlier work intentionally scoped itself to Running and Attention; this follow-up expanded the product scope. Existing architecture-refactor guidance already requires entry-point inventories, and regression guidance already requires cross-feature journeys for neighboring commands.

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -812,7 +812,7 @@
   {
     "id": "AI-SESSION-QUICK-SWITCH-COMMANDS-001",
     "domain": "session",
-    "title": "The Switch to AI Session picker lists every live local session most-recently-focused first plus one entry per other window with running sessions, jumping locally with terminal focus and conversation open or handing off through the focus mailbox with plain-navigation degradation, while Toggle Last AI Session alternates the two most recently focused local sessions from a terminal-focus-fed MRU that prunes dead sessions, and both commands are contributed without default keybindings",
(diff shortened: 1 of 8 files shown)


hzcheng:
lgtm

PR #202, opened 2026-08-10
Fix stale attention projection during session switching

hzcheng:
## Summary
- keep the dedicated attention projection authoritative across asynchronous workspace HTML replacement
- prevent cross-window acknowledged completion events from being replayed by the local recovery projection
- add browser and controller regressions for stale attention flashes and aggregate authority

## Verification
- npm run test:ci:linux
- manual multi-window Next Attention Session verification
- installed VSIX byte verification against the active VS Code Server

## Skill harvest
No skill change was justified. The regression workflow already requires a RED user-visible test, final verification, and review; the remaining lesson is product-specific projection protocol design and is covered by the follow-up architecture refactor.

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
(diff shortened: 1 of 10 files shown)


hzcheng:
lgtm

PR #203, opened 2026-08-10
Fix session runtime continuity and conversation recovery

hzcheng:
## Summary

- Preserve active AI session actions when workspace roots are added or removed.
- Restore attention lookup, focus ownership, pending projection, and running counts across workspace scope changes.
- Fail closed when old and new scopes contain ambiguous runtimes for the same session.
- Retry one transient authoritative Conversation read before reporting it unavailable.
- Accept valid large Codex thread/read responses while retaining a bounded transport limit.
- Prevent a running Active Session card from showing stale Attention through either authoritative rendering or an incremental Webview update.
- Keep release VSIX assets synchronized with their source files.

## Root causes

Workspace root changes create a new scope identity while existing runtimes retain their original identity. Session hydration already accepted navigation and root continuity, but command and attention paths still required exact scope equality. The sidebar could therefore show an Active session that the host rejected as inactive, and attention stopped resolving the same runtime.

Conversation opening had two independent failure modes. Transient app-server failures were surfaced without a bounded retry. After that retry was added, the real current Codex 0.147.0 thread/read response had grown to 17.25 MiB and was rejected by the legacy 16 MiB frame limit before JSON parsing.

The running-card regression had two paths. Host hydration independently combined a newer running execution snapshot with an older Attention aggregate, and the direct Webview Attention stream reapplied stale event IDs without checking the rendered execution state. Existing controller tests did not own the final rendered race, while older browser Attention fixtures incorrectly modeled sessions as running while expecting red dots.

## Architecture

One shared workspace runtime continuity rule is reused across hydration, terminal commands, attention handling, and dashboard running counts. Exact backend identity operations remain strict; continuity is limited to workspace ownership and candidate resolution.

Conversation recovery stays bounded: only unavailable reads are retried, once. Unknown sessions, empty conversations, superseded intents, and persistent failures keep their existing outcomes. Codex thread/read accepts frames up to the existing 64 MiB conversation-source safety boundary and continues to reject larger or unterminated responses before parsing.

The Active Session projection now enforces the user-visible invariant that running and Attention are mutually exclusive. Incremental DOM reconciliation applies the same invariant without deleting the authoritative event, so the red dot can appear after the session actually stops.

## Testing

- npm run test:ci:linux
- Unit tests: 704 passed
- Contract tests: 873 passed
- Integration tests: 397 passed
- Browser tests: 180 passed
- npm run test:release-packaging
- npm run test:architecture-guards
- Changed-line coverage: 97.73 percent
- Compiled production client read the real 17.25 MiB Codex 0.147.0 thread/read response with 58 turns and no diagnostics; no conversation content was logged
- Running/Attention exclusivity is covered at Host-rendered and live browser-interaction surfaces

## Skill harvest

none — no additional skill change was justified for the running/Attention follow-up because fixing-regressions-with-ci already requires RED first, final rendered or interaction ownership, and verification after the final fix. Earlier work in this PR updated .skills/developing-provider-conversation-adapters for framed-response byte and duration probes.

### DIFF
--- a/.skills/developing-provider-conversation-adapters/SKILL.md
+++ b/.skills/developing-provider-conversation-adapters/SKILL.md
@@ -21,6 +21,9 @@ Green self-authored fixtures alone are not evidence.
    docs, or a sibling provider's layout:
    - census record shapes: `type` distribution, field presence per record
      kind, single-record size bounds
+   - for a framed protocol, record the sanitized full response byte size and
+     duration of a valid large live response, verify it remains below a
+     bounded transport cap, and never infer response size from source-file size
    - find lifecycle markers (start / finish / interrupt) and cross-file
      links (toolUseId, promptId, sidechain flags)
    - detect continuation records (distinct promptIds, injected user turns)
--- a/docs/superpowers/specs/2026-07-25-active-session-conversation-outline-design.md
+++ b/docs/superpowers/specs/2026-07-25-active-session-conversation-outline-design.md
@@ -728,7 +728,7 @@ authoritative update.
   payload. The viewer retains at most 100 interactions or 4 MiB, whichever is
   reached first, evicting the farthest page while preserving the selected
   anchor.
-- One Codex app-server response is capped at 16 MiB and every request has a
+- One Codex app-server response is capped at 64 MiB and every request has a
   ten-second deadline.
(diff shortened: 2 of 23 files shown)


hzcheng:
lgtm

hzcheng:
lgtm

PR #206, opened 2026-08-10
Fix lingering Extension Host workers after successful tests

hzcheng:
## Summary

- add a versioned IPC completion signal that is sent only after the real Extension Host test succeeds
- let the parent watchdog terminate a completed worker that retains cleanup handles without masking failures or unknown messages
- add deterministic launcher, runner, and worker coverage plus a CI-owned behavior contract

## Verification

- `npm run test-compile`
- focused Extension Host tooling tests (15 passed)
- `npm run test:ci:linux` (changed-line coverage 91.11%)
- `xvfb-run -a npm run test:extension-host` with VS Code 1.130.0; the real smoke passed and exited immediately
- `git diff --check`

## Skill harvest

None. The existing regression/CI skill already requires reproducing and tracing the root cause; the earlier ambiguity was a compliance issue rather than missing reusable guidance.

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -3562,6 +3562,23 @@
     ]
   },
   {
+    "id": "RELEASE-EXTENSION-HOST-WORKER-COMPLETION-001",
+    "domain": "release",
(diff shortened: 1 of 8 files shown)


hzcheng:
lgtm

PR #207, opened 2026-08-10
Prevent starvation in running session rotation

hzcheng:
## Summary

- Separate cross-window rotation state from each window's local-session cursor so a multi-session window cannot permanently skip one running session.
- Carry source, target, and creation time through running-focus protocol v3 to preserve a stable ring anchor and reject stale handoffs that would reverse newer navigation.
- Cover multi-session windows, degraded three- and four-window rings, and late handoff ordering with CI-owned regression tests.
- Version the separately packaged Attention UI Bridge as 1.0.2 so existing installations can receive protocol v3.

## Scope

This change guarantees that every stable running session and running window remains reachable within finite navigation steps. It does not define a strict human-visible A-B-C window order; that requires the planned rotation state-ownership refactor.

## Skill harvest

updated .skills/publishing-and-merging-github-prs — protocol v3 exposed that separately packaged companion runtime changes must raise their manifest version before publish; quick validation and repository skill owner tests pass.

## Verification

- npm run test-compile
- 53 focused running-focus, attention-focus, and bridge contract tests
- npm run test:architecture-guards
- brand identity and repository skill owner tests
- npm run test:release-packaging
- git diff --check
- Manual: current starvation scenario is usable; strict A-B-C ordering remains explicitly out of scope


hzcheng:
lgtm

PR #213, opened 2026-08-11
Refactor attention acknowledgements into correlated transactions

hzcheng:
## Summary

- Replace fire-and-forget card acknowledgements with a versioned, request-correlated transaction.
- Resolve the complete attention owner from current Host workspace authority and reject stale, partial, malformed, or conflicting requests.
- Keep Webview pending state until a committed authoritative v3 presentation is applied, with bounded failure recovery and workspace-scope isolation.
- Make bridge persistence outcomes explicit and retain idempotent Host settlements for duplicate delivery.

## Testing

- npm run test:ci:linux
- Focused attention transaction suite: 56/56
- Repeated workspace-scope and production v3 browser paths: 20/20
- Changed-line coverage: 97.79% (177/181)

## Review

Independent protocol and test reviews found no remaining Critical or Important findings. Final merge-base-to-HEAD integration review passed.

## Skill harvest

No skill change was justified. The only reusable workflow failure was overlapping commands that clean and consume the shared out directory; review-fix-commit-loop already explicitly prohibits this scheduling pattern. The full Linux gate subsequently passed when run in dependency order.

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -4109,14 +4109,18 @@
   {
     "id": "ATTENTION-SESSION-CARD-ACKNOWLEDGEMENT-001",
     "domain": "attention",
-    "title": "Clicking a stopped Active Session card acknowledges every authoritative owner event and clears all Dashboard attention surfaces without stale bridge resurrection",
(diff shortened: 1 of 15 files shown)


hzcheng:
lgtm

PR #214, opened 2026-08-11
feat: add global session status dots to the AI Conversation header

hzcheng:
## Summary

Adds a pair of global Session status dots to the AI Conversation header, between the identity and the navigation buttons:

- **Green dot: running Sessions** — summed from every window's `runningAiSessionCount` in the open-workspaces aggregate, so the count is cross-window by construction.
- **Red dot: Sessions needing attention** — `getEffectiveAggregate().sessions.length`, the same global (local + remote bridge) figure the status bar already shows.

Each dot shows its count beside it, pulses while its count is non-zero, dims with a `0` when idle, explains itself in a tooltip, and honors `prefers-reduced-motion`. The identity and navigation sides of the header are now equal-width flex containers so the status group sits at the true center of the row.

Host side, a small `ConversationSessionStatusController` embeds the current status into the rendered document (plus a `data-session-status-request-id` watermark) and pushes correlated `conversation-viewer-session-status` messages whenever either aggregate changes, deduplicating unchanged statuses and rebuilding the document after a failed delivery — the same pattern as the telemetry channel. Review hardening: after target transitions (retarget, in-place subagent switch, resume-from-suspend) the viewer republishes the status to heal Webview-side guard discards that would otherwise poison the dedup and leave stale dots; host-side counts clamp to the Webview validator bound.

The dormant (third) dot is deliberately deferred: broadcasting idle state needs an open-workspaces bridge protocol extension and a lockstep bridge VSIX release.

Behavior: `CONVERSATION-SESSION-STATUS-001` (unit + contract + integration + browser owners), assigned to `MAIN-AI-SESSION-CONVERSATION-OUTLINE`.

## Skill harvest

updated .skills/resilient-webview-mutation-protocols — two conversation Webview harness lessons from this branch: version-skew strips must attach to the correct per-script fixture chain (verified by re-executing the chain against the pinned sha256), and host-rendered browser documents only receive CSS with `includeStyles: true` + a theme fixture (CSP blocks `addStyleTag`). Validated with `run-skill-management-checks.js` and `tests/unit/tooling/repositorySkills.test.js` (10/10); recorded as the `Skill-Harvest` trailer of the capability audit commit.

## Verification

- `npm run test-compile`; unit 6/6 (`conversationSessionStatusController`), contract `projection` (running-count sum), integration 84/84, browser 95/95 (incl. new rendering/centering/reduced-motion/stale-message assertions and the byte-exact version-skew fixture)
- `npm run test:behavior-contracts` (catalog + capability audit currency), architecture guards 82/82, ai-session-safety checks, dashboard Webview checks, `npm run test:release-packaging`, `npm run lint`, `git diff --check`
- Rendered-pixel screenshots at 700px and 240px for lit and dim states; header centering asserted ≤2px in the browser suite
- Local install (`SKIP_NPM_CI=1 npm run install-local`) after rebasing onto `origin/main` (`ec590579`), with VSIX↔installed SHA-256 comparison of `dist/dashboard.js`, `media/conversationViewerScripts.js`, and `media/conversationViewer.css`
- In-conversation manual trial of the installed feature build drove the counts-beside-dots and true-centering iteration


hzcheng:
lgtm

PR #222, opened 2026-08-11
Fix focused session card continuity across workspace updates

hzcheng:
## Summary

- reject AI Session presentations that belong to a different rendered workspace before adopting their revision or focus state
- validate atomic AI and OPEN workspace replacements both before mutation and against the resulting DOM, including reused semantic-revision no-op updates
- preserve the focused card highlight and focused-card Conversation action while requesting an authoritative refresh for mismatched updates
- strengthen architecture mutation guards for workspace-scoped Presentation adoption

## Verification

- `npm run test:ci:linux`
- focused Chromium coverage for focused-card highlight, reused OPEN semantic revisions, and focused-card Conversation opening
- source/media byte identity and `git diff --check`

## Skill harvest

Updated `.skills/installing-vscode-extensions-locally`: verified installed bytes can still be inactive when the Extension Host predates the install, so the workflow now requires `Developer: Reload Window` and post-reload activation evidence before behavioral testing. Validated with `quick_validate.py` and the repository skill owner tests.

### DIFF
--- a/.skills/installing-vscode-extensions-locally/SKILL.md
+++ b/.skills/installing-vscode-extensions-locally/SKILL.md
@@ -96,6 +96,20 @@ Build and install the Agent Pivot extension that matches the current VS Code env
      establish only packaging or attempted installation, never verified bytes.
    - if a script exits 0 with warnings, report warnings separately from failure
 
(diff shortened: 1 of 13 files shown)


hzcheng:
lgtm

PR #223, opened 2026-08-11
Fix OPEN dashboard refresh storms

hzcheng:
## Summary
- align Session Presentation navigation identity with the current workspace card rendered from UI Bridge registration
- keep live Extension Host workspace scope as the session authority
- avoid full card hydration on direct focus publication and add architecture guards
- cover mismatch-to-full-document convergence so refresh requests cannot self-amplify

## Root cause
The OPEN card used the UI-host normalized navigation identity while Session Presentation used the local Extension Host identity. During registration convergence, the Webview rejected otherwise valid updates and repeatedly requested full refreshes, causing card and page flicker.

## Verification
- `npm run test:ci:linux`
- focused OPEN convergence contract and browser owners
- 153 architecture guard tests
- Dashboard Webview, AI session safety, and OPEN workspace safety checks
- independent exact-SHA integration review: PASS

## Skill harvest
- none: existing regression, resilient Webview mutation, review, installation, and publication skills already covered the reusable workflow; no instruction gap was found.

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -5407,6 +5407,22 @@
     ]
   },
   {
+    "id": "OPEN-WORKSPACE-PRESENTATION-CONVERGENCE-001",
+    "domain": "webview",
+    "title": "OPEN transactions align Session Presentation with the rendered current workspace identity while UI-host registration converges",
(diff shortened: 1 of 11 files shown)


hzcheng:
lgtm

PR #224, opened 2026-08-11
perf: make AI conversation session switching fast

hzcheng:
## Summary

Speeds up AI Conversation session switching (card click, terminal/window focus, previous/next session, running/attention jumps) by removing redundant work from every hop of the switch pipeline:

- **Provider layer**: Claude `resolveConversationSource` reused a full synchronous `~/.claude/projects` tree scan per conversation load; it now resolves through the session-file index, armed only by unbounded scans and bound to a cheap projects-tree signature (subdirectory names + mtimes), so any transcript appearing/moving/being deleted forces an immediate rescan and the decline-on-duplicate boundary is preserved with no staleness window. The Codex large-thread cache was cleared on *every* provider watch event; invalidation now runs once per 250ms debounce window, right before subscribers are notified (and a canceled debounce still invalidates on watch release).
- **Host render layer**: every publication re-rendered the whole retained window (≤100 interactions / 4 MiB) through markdown-it + highlight.js — including 1Hz streaming refreshes and pure selection changes. A per-message render cache keyed by deterministic message id + render signature (session, thinking visibility, response state, computed clock text) makes publication cost proportional to what changed; retention-layer replacement invalidates by interaction id, and each entry carries a render version. Warmup (adjacent-session prefetch + warm-snapshot revalidation) is serialized instead of racing the switch.
- **Protocol layer**: publications carry an `htmlSignature` from a collision-free registry (exact render stream → monotonically minted token, never a hash). The Host omits the HTML string only when the Webview's correlated `conversation-viewer-applied` ack (generation + requestId + signature) proves that exact content is applied; until then every publication carries full HTML, so lost/unapplied pages are always redelivered. Unapplyable deltas trigger a bounded full resync (one rebuild per publication).
- **Webview layer**: delta pages skip DOMPurify and the DOM reconcile entirely; a small per-session LRU of sanitized pages makes switching back to a recent session skip the multi-megabyte sanitize pass.

## Skill harvest

updated .skills/resilient-webview-mutation-protocols — two Common Mistakes rows from this branch's review rounds: omission signatures must carry content identity (render version), and `postMessage` delivery is not application (only a correlated applied ack may authorize omission; unapplyable deltas must request a bounded resync). quick_validate.py is not available in this environment; validated via the branch's contract/browser suites instead.

## Verification

- `npm run test-compile`; unit 809/809, contract 890/890, integration 406/406, browser 216/216 (all `node --test` suites, fresh after the final fix and rebase onto `de8cf075`).
- `scripts/run-conversation-performance-checks.js` (budgets in `.ci/conversation-performance.json`): host session switch 8.0ms → 2.2-3.1ms, host initial publication 17.9ms → 12.0-12.8ms, host incremental refresh 5.7ms → 4.1-4.4ms on the CI fixture; larger sessions benefit more (render cost now scales with changed messages, not the 4 MiB window).
- `scripts/check-behavior-contracts.js`, `run-dashboard-webview-checks.js`, `run-ai-session-safety-checks.js`, `run-architecture-guards.js`, `test:release-packaging`, `npm run lint`, `git diff --check` — all exit 0.
- New coverage: render-cache/registry unit tests, delta ack-gating integration tests (no ack → full HTML; wrong ack → full HTML; correlated ack → delta; content change → full HTML with new token), webview delta/sanitize-cache/resync browser tests, Claude duplicate-decline contract tests, Codex debounce-window cache tests.
- Two read-only review passes over the full merge-base diff; all Critical/Important findings fixed and re-verified.


hzcheng:
lgtm

PR #225, opened 2026-08-11
perf: restore conversation frames on session switch-back

hzcheng:
## Summary

Round two of conversation-switching fluency: repeat switches (MRU toggle, previous/next round-trips, clicking back to a card) now reattach the already-built DOM instead of transferring, sanitizing, parsing, and rebuilding megabytes of HTML.

- **Webview frame cache**: on a session switch, the live conversation is stashed as a detached frame (DOM nodes, reconcile signatures, worklog expansion, reading anchor, scroll position) bounded by 4 frames **and** a total node budget, with mermaid blob URLs released on eviction. When an incoming page's content token matches a stashed frame, the same DOM nodes are reattached — no sanitize, no parse, no reconcile.
- **Webview-authoritative frame inventory**: every applied ack carries the Webview's current frame list, and the Host offers `restoreFrame` (HTML omitted) only for tokens on that list — never inferred from past acks. An evicted frame gets full HTML; a false offer ends in one bounded resync, and a rebuilt document reports an empty inventory, so no rebuild cascade is possible.
- **Navigation intent wins over restored position**: a frame restore carrying a new `selectedInteractionId` centers that target (or scrolls to the latest) instead of resurrecting the stashed scroll position.
- **Token stability**: the render cache keys entries by session id + message id (colliding provider ids no longer thrash versions), and retention invalidation marks entries stale — a re-render producing byte-identical HTML keeps its version, so unchanged content reloads keep their token while changed content always mints a fresh one.
- **Mermaid**: `releaseExcept` spares stashed frames' blob URLs from global releases; pending renders reset at stash and re-render on restore.

## Skill harvest

updated .skills/resilient-webview-mutation-protocols — the version-skew strip row now prescribes the byte-exact verification method used here (apply candidate strips, diff byte-for-byte against `git show <base>:<script>`) instead of slower hash guessing. Validated through the branch's contract/browser suites (quick_validate.py unavailable in this environment).

## Verification

- Fresh after the final fix: unit 810/810, contract 890/890, integration 408/408, browser 221/221; `run-conversation-performance-checks.js`, `check-behavior-contracts.js`, `lint`, `run-dashboard-webview-checks.js`, `test:release-packaging`, `git diff --check` — all exit 0.
- New budget `webviewWarmFrameRestoreMs: 100` in `.ci/conversation-performance.json` — warm restore measures ~13ms for the 100-message fixture (the initial full application budgets 500ms).
- New coverage: frame restore node identity + zero-sanitize + scroll restoration, fresh-navigation-target centering, six-session churn eviction + resync behavior on both protocol sides, ack-gated `restoreFrame` including "changed content never gets a restore", render-cache stale/revive and composite-key unit tests, mermaid `releaseExcept` spare/evict test, `conversation-viewer-applied` frames validation.
- Read-only review of the merge-base diff: no Critical/Important findings; both Minor gaps it flagged are now pinned by tests.


hzcheng:
lgtm

PR #226, opened 2026-08-11
fix: correlate conversation resync requests and announce reused-panel session switches

hzcheng:
## Summary

Fixes the stuck old-conversation content observed when switching into a large (Codex) session — two stacked defects, plus the loading gap that made the wait invisible.

- `conversation-viewer-request-sync` now carries the subscription generation and the `{projectId, provider, sessionId}` identity of the page the Webview failed to apply. The Host rebuilds the document only while it still owns that exact generation and session; requests stranded by a newer switch are ignored because the incoming target's own delivery and ack closure owns the recovery.
- The Webview resync gate is per generation instead of per document: a rapid A→B→C switch that misses the frame restore twice now escalates each miss exactly once, instead of consuming the single document-level gate on the first miss and stranding the final session on the outgoing DOM forever.
- Reusing the panel for a different session posts a lightweight `conversation-viewer-loading` notice: the outgoing conversation stays visible, dimmed (`opacity: .55`) and `aria-busy`, with `Loading conversation…` announced in the status line, until the incoming session's first publication lands. A fresh panel keeps rendering its own loading document; same-session reloads never dim live content.

## Root cause

A resync request had no correlation identity. When it arrived after `replaceTarget` had cleared the superseded `latestPublication`, the Host dropped it silently; the next publication could still be HTML-less (`restoreFrame` offered from the last reported frame inventory), and when that restore also missed — common with large sessions because the stash evicts the least-recent frame under the node budget — the Webview's one-per-document gate had already been consumed, so no further escalation was possible. Stopped sessions publish nothing more, so the wrong DOM persisted indefinitely.

## Tests

- Browser: rapid A→B(miss)→C(miss) switch posts two correlated resyncs and the final DOM is C after the full republication; loading notice arms/clears around the incoming first publication; stale and malformed notices are inert. Existing resync assertions updated to the correlated shape.
- Integration: stale generation/session resyncs are ignored (each correlation check isolated as its own sensitive mutation, proven by temporarily reintroducing each defect); a correlated resync in the first-publication load gap is safely ignored and the in-flight load still delivers full HTML; reused-panel switches post exactly one notice before the incoming publication, fresh panels and same-session reloads post none.
- Unit: protocol validator accepts the exact 6-field shape and rejects missing/extra/mistyped fields.
- `previousViewerScript` strip chain extended for every Webview hunk; standalone verification confirms byte-exact equality with the pinned fixture hash before the full browser suite ran.

## Verification

- `npm run test:ci:linux` exit 0 on the final head (unit 810 / contract 890 / integration 412 / browser 105, lint, safety, performance budgets, release packaging).
- Read-only review of the merge-base diff: no Critical/Important findings; follow-up commit pins the review-called-out load-gap invariant.

## Skill harvest

Reviewed task friction per `harvesting-workflow-lessons`: the strip-chain standalone-verification method, the mutation-sensitivity practice, and the exit-code-piping pitfall all already live in the existing skills and worked as documented. No skill change justified — recorded as `Skill-Harvest: none` in the capability audit commit.


hzcheng:
lgtm

PR #227, opened 2026-08-12
perf: validate the Codex conversation cache by rollout stat instead of TTL

hzcheng:
## Summary

Codex conversation content can only be read through a full `thread/read` (includeTurns) RPC — measured ~3.1s for a 50MB / 205-turn session. The adapter's large-conversation cache previously expired every 15s and was discarded wholesale on every provider invalidation, so switching back to a big stopped session — or any unrelated session activity in the store — forced the full read again even when the content provably had not changed.

The cache is now **validated by the rollout's stat signature** instead of a TTL, with explicit memory bounds and cost-based eligibility:

- Each entry stores the rollout file's `${size}:${mtimeMs}` signature, captured before the provider read. A read hits only when a fresh probe returns the identical signature — an unchanged stat means the app-server content cannot have changed, which is exactly the assurance level the session watch already polls (`id:path:size:mtime` diff). A signature captured mid-read can never serve newer content than it claims: any append invalidates the next probe.
- **Eligibility matches real read cost**: cached when visible text ≥ 512KiB *or* the full read took ≥ 100ms — tool-record-heavy threads with little visible text (real reads of ~350–730ms) are covered too.
- **Memory is explicitly bounded**: 2 entries max, a shared 8Mi-character budget (evict oldest, keep newest), a ten-minute idle sweep on reads, immediate release when an entry's stat mismatches or becomes unreadable, and deterministic release on `dispose()`. A viewer-idle release hook was evaluated and rejected: session switches momentarily empty the watch subscription set, so it would have cleared the cache on every switch.
- `resolveSessionFilePath` revalidates the lifecycle path cache with a single `existsSync` instead of re-scanning the whole sessions tree per call, so each probe costs one `existsSync` + one `statSync`.
- Conversation content itself remains app-server-only by design — the stat is a cache-validity signal, never a content source.

Measured effect (synthetic 205-turn / ~12MiB-normalized scenario): full read 105.5ms → stat-validated reload **0.051ms**; the stat-changed full reload is now budgeted (`codexAppendReloadMs: 500`) instead of merely counted. Warm switches then compose with the webview frame cache (~13ms restore) end to end.

Honest scope: this is memoization, not incremental transfer — an actively appending large session still pays a full read per change (cost grows with history). Closing that needs the app-server's paginated/item-level reads (`thread/turns/list` / `thread/items/list`, binary-string evidence in codex 0.147.0), which stay unprobed and out of scope here. The `size+mtimeMs` signature cannot detect a same-size same-mtime file replacement (inode/boundary-hash hardening noted as follow-up, non-blocking).

## Tests

- Contract tests rewritten from TTL semantics to stat semantics: unchanged stat ⇒ cached indefinitely (the harness's synchronous timers would fire any hidden expiry); changed stat ⇒ fresh read; unreadable/throwing probe ⇒ bypassed and released; invalidations retain stat-valid entries; watch release retains them; in-flight invalidation still prevents caching; two-entry FIFO, 8Mi-character budget eviction, ten-minute idle eviction, and duration-based eligibility each pinned. Probe receives the native (subagent-split) thread id.
- New probe module unit test (size+mtime signature, missing/dir/empty paths) and `resolveSessionFilePath` revalidation test (deleted rollout ⇒ null; recreated ⇒ resolves).
- Perf harness gains two Codex scenarios (large shape; tool-heavy shape pinned to the duration gate via an injected clock) with budget keys `codexStatValidatedReloadMs: 50` / `codexAppendReloadMs: 500`, plus cached-character and counter-invariant assertions.

## Real-data verification (per adapter skill)

Compiled adapter against the live codex app-server (0.147.0) and real on-disk rollouts, read-only:

- Small session: outline interaction count matches the raw `thread/read` user-turn ground truth; fast small reads correctly stay uncached.
- Large session (59.9MB rollout, 502 interactions): first read 1 RPC; second read **0 RPCs**, byte-identical output; forged stat forces a fresh read with an identical content revision.
- Tool-heavy session (36MB rollout, 200 interactions, sub-gate visible text): first read 730ms; second read **0ms, 0 extra RPCs** — the duration gate covers the exact shape the old char gate missed.
- 175MiB live session: read succeeds (its sanitized response fits the 64MiB frame cap); no hang or failure-path change.

## Review

Two read-only review rounds plus the owner's review. Fixed along the way: probe paid a full recursive sessions-tree scan per read (now the lifecycle path cache fast path); then the owner's two Importants — unbounded cache memory (now byte budget + idle sweep + immediate stale release) and eligibility/perf gates not reflecting real read cost (now duration-based eligibility + append-reload and tool-heavy budgets). Final re-review: no Critical/Important.

## Skill harvest

`updated:.skills/review-fix-commit-loop` (multi-edit replace calls report replacements, not matched edits) — recorded in the first audit commit trailer; the bounds follow-up round recorded `none`.

### DIFF
--- a/.ci/conversation-performance.json
+++ b/.ci/conversation-performance.json
@@ -2,6 +2,7 @@
   "adapterColdOutlineMs": 1500,
   "adapterCachedOutlineMs": 100,
   "adapterAppendMs": 250,
+  "codexStatValidatedReloadMs": 50,
   "hostInitialPublicationMs": 250,
   "hostIncrementalRefreshMs": 150,
   "hostSessionSwitchMs": 250,
--- a/.ci/conversation-performance.json
+++ b/.ci/conversation-performance.json
@@ -3,6 +3,7 @@
   "adapterCachedOutlineMs": 100,
   "adapterAppendMs": 250,
   "codexStatValidatedReloadMs": 50,
+  "codexAppendReloadMs": 500,
   "hostInitialPublicationMs": 250,
   "hostIncrementalRefreshMs": 150,
   "hostSessionSwitchMs": 250,
--- a/.skills/review-fix-commit-loop/SKILL.md
+++ b/.skills/review-fix-commit-loop/SKILL.md
@@ -120,6 +120,11 @@ Summarize:
   selectors before trusting tests: a replacement anchor like `.foo {` also
   matches inside a longer selector such as `.bar .foo {`, and Sass happily
   compiles the mangled nesting.
+- A multi-edit replace call reports applied replacements, not matched
+  edits: an edit whose anchor does not match is silently skipped and only
+  shows up as a lower replacement count than the edit count. Compare the
+  two numbers and grep the anchors before running tests, or an unapplied
+  hunk rides into the commit unnoticed.
 - Adding a Dashboard command or an activation-time `vscode` API (e.g.
   `createStatusBarItem`) fails activation-harness and integration tests with
   opaque `waitFor` timeouts far from the cause: bootstrap errors are swallowed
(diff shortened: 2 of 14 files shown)


hzcheng:
lgtm

PR #228, opened 2026-08-12
perf: reload large Codex conversations incrementally via thread/turns/list

hzcheng:
## Summary

Large Codex conversations previously paid a full `thread/read` on every content change (~1s at 60MB; the 175MB session cannot even finish within the 10s request timeout). This PR reloads cached root-thread conversations incrementally through the experimental `thread/turns/list` tail page, re-normalizing only the turns that changed.

- Client: opts into `capabilities.experimentalApi` at initialize and reports the server version. 0.147+ omits `serverInfo` from the initialize response, so the version is parsed from the `userAgent` product token (`<originator>/<major.minor.patch> …`).
- Adapter: normalization splits into per-turn chunks behind one shared code path, and the source revision is composed from per-turn fingerprints — full and incremental loads of identical content produce byte-identical output and identical revisions.
- Accelerator discipline, since the method is experimental: a version allowlist (0.147 only, re-probed in the included spike), a process-lifetime circuit breaker on method rejections/malformed pages (transient `unavailable`/`timeout` transport errors deliberately do **not** trip it), and subagent threads always stay on the stable full read (their seeded dispatch interaction accumulates across turns).
- The anchor walk is cost-aware: a slow first page (≥250ms = the legacy rollout-replay backend) exits to a full read after one probe, fast pages walk within a 1024-turn/3s budget, and a tail rewritten by compaction/rollback on the indexed backend is **rebuilt from the fetched pages** — a full `thread/read` of a 175MB-class session would exceed the 10s request timeout outright.
- The rollout lifecycle is re-applied on every read (cache entries hold base interactions only): a session that stops while its visible content is unchanged still settles its Working indicator, even when the tail page re-anchors the cache without re-reading content.
- Perf: the synthetic 12MiB append reload drops ~96ms → ~2ms, and the `codexAppendReloadMs` budget tightens 500ms → 50ms.

Known limits: cold (first) loads still use `thread/read`, so a 175MB-class session still exceeds the 10s request timeout on first load (cold-path paging via `thread/resume.initialTurnsPage` is documented in the spike as a separate, larger experimental surface). Legacy pre-index sessions pay a server-side full rollout replay per tail page — never worse than the status quo, and the stat-signature cache still covers the unchanged case.

## Skill harvest

updated .skills/developing-provider-conversation-adapters — the Codex row now covers the `thread/turns/list` incremental surface, and a new note records the 0.147 initialize lesson (no `serverInfo`; parse the version from `userAgent`) that real-data validation caught while stub tests stayed green.

## Verification

- `npm run test:ci:linux` exit 0 (includes unit 811 / contract+integration 1315, architecture guards, behavior contracts, capability audit currency).
- Focused: `codexConversationAdapter.test.js` 71/71, `codexAppServerClient.test.js` 28/28; perf script green (`codexAppendReloadMs` ~1.7ms vs 50ms budget).
- Mutation-tested: anchor-miss rebuild, slow-page early exit, empty-page-with-cursor fallback, circuit breaker, lifecycle re-application, and cross-path revision equality assertions each fail when the corresponding source line is reverted.
- Real-data validation against codex 0.147.0 and real sessions (compiled `out/` adapter, real app-server):
  - live thread: a real `turn/start` append reloads via `thread/turns/list` only and deep-equals a fresh full-read outline (identical revision);
  - legacy 60MB session: stat flip with unchanged content re-anchors in ~614ms vs ~1023ms full read (server-side replay bound);
  - paginated 10.9MB session: same re-anchor in ~7ms vs ~443ms;
  - capability-less client: a real `-32600` rejection trips the breaker, falls back, and sticks.
- The probe behind the protocol claims is committed under `spikes/codex-paginated-read/`.

### DIFF
--- a/.ci/conversation-performance.json
+++ b/.ci/conversation-performance.json
@@ -3,7 +3,7 @@
   "adapterCachedOutlineMs": 100,
   "adapterAppendMs": 250,
   "codexStatValidatedReloadMs": 50,
-  "codexAppendReloadMs": 500,
+  "codexAppendReloadMs": 50,
   "hostInitialPublicationMs": 250,
   "hostIncrementalRefreshMs": 150,
   "hostSessionSwitchMs": 250,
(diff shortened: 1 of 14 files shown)


hzcheng:
lgtm

PR #229, opened 2026-08-12
perf: window Codex cold starts behind summary skeletons

hzcheng:
## Summary

Phase 3 of the Codex large-session loading work (after #227 stat-signature cache and #228 incremental paginated reload): **the cold start**. A 183MB paginated session's first `thread/read` takes 40.6s / 52.4MB and exceeds the 10s request timeout — today it cannot be opened at all.

On verified 0.147 paginated backends, the cold start now:

1. awaits the completed handshake (`client.ensureReady()`) before version gating — a sync gate can never engage on the first read;
2. walks **all** turns as summary skeletons (`itemsView:"summary"`, ~35ms/50-turn page, flat on the indexed backend) — outline-complete, so `isStart`/`totalInteractions` keep their exact local meaning;
3. materializes full items only for the tail window (~40 turns / 1.5MB raw);
4. materializes older windows on demand inside `readPage`/`readSnapshot` by replaying the walk's recorded boundary cursors with `itemsView:"full"` (probe-verified portable across limit/itemsView, byte-identical items).

Design invariants (four review rounds encoded):

- **Projection revision**: `sha256(content-epoch stat | per-turn summary fingerprints | structure generation)`. Any observed provider content change moves it (the viewer skips refreshes only when revision AND ids AND responseStates are unchanged — a tool-only append must still republish); pure hydration never moves it, so retained-page cursors survive scrolling.
- **Summary fingerprints exclude the agentMessage**: real-data census (218 turns) found the summary agent text unreliable — omitted for interrupted turns, divergent in one completed turn. The user side is verbatim everywhere.
- **Skeleton ids == full-path ids** (both are the first userMessage item id), so anchors survive materialization; a multi-userMessage turn (exists in real data) expands 1→N on materialization, bumps the structure generation, and moves the revision so cursors anchored at new ids are never left dangling across cache eviction.
- **Freshness closure**: on a stat move, every materialized chunk the incremental anchor walk does not re-verify demotes back to a skeleton — in-place historical edits (the protocol's `updated_at_ordinal` makes them real) can never serve stale content; the next view re-materializes live. Materialization commits re-verify the entry's cache slot AND stat signature after every landed RPC (the last-page race is closed), and the cold-start walk/tail close over their signature with one re-walk retry.
- **Materialized-window LRU**: a per-entry 4Mi budget demotes least-recently-served full chunks outside the protected ranges (served window + tail zone) back to skeletons — long scrolling cannot re-inflate an entry toward the whole conversation.
- **Legacy verdict**: rolling median page latency (up to 3 pages, min 2; handshake excluded). A fast first page no longer suppresses the verdict; a single jittery page never triggers it. Verdict → the stable full read (no size-based feasibility guessing); if that read times out / is too large, the forced slow walk still opens the session — huge legacy sessions go from "never opens" to "opens slowly". Runaway valves (4k pages / 120s) are resource bounds, not heuristics.
- **Live-session hazard** (found by real-data validation): item ids are not reliably turn-scoped while a session is being written — a response-spanning reasoning item was projected into different turns by fetches seconds apart. Cross-page id collisions invalidate the entry and re-walk a consistent snapshot; they never circuit-break the accelerator.

Deliberate trade-offs, called out honestly:

- The outline folds multi-userMessage turns to their first message until materialized (the server projection exposes only `first_user_item_id`); it self-heals on the next content change. Measured divergence on the biggest real session: exactly one turn, two folded entries.
- 60MB-class legacy cold starts cost ~2.1-2.5s instead of ~1.0s (verdict pages before the full read). Bounded and shrinking (legacy = pre-0.147 sessions); the alternative is misjudging paginated sessions into a guaranteed timeout.
- A stat move with materialized history beyond the anchor now moves the revision (demotion) — fake-out zero-churn is preserved only when nothing unverified remains materialized. Re-materialization is lazy and the viewer's HTML signature suppresses redundant publications, so the user-visible cost is nil; the RPC cost is a few small pages.
- The windowed revision basis differs from the full-read basis by construction; crossing bases is a one-time cursor reset on the standard stale-recovery path.

Small sessions (<4MiB rollout), subagent threads, unverified server versions, and signature-less environments all keep the previous full-read path byte-identical.

## Skill harvest

Updated `.skills/developing-provider-conversation-adapters` with probe-verified protocol facts: summary = first userMessage (+ unreliable final agentMessage), cursor portability across limit/itemsView, handshake-first version gating (sync gates are circular at cold start), handshake contamination of first-page latency, non-turn-scoped item ids on live sessions, and the viewer's revision+ids+responseState refresh-skip rule that forces revisions to move on any content change.

## Verification

- **Real data, production client** (codex 0.147.0, live 183MB paginated session): cold start **~420ms** (was 40.6s / guaranteed timeout); outline equal to the full-`itemsView:"full"`-walk ground truth except the known folded turn; tail page and a deep materialized page **byte-identical** to ground truth; revision stable across quiet hydration; expansion of the folded turn moves the revision as designed. 60MB legacy: ~2.1-2.5s total, outline exact. Telemetry `thread/read includeTurns:false` on the 183MB session: ~250ms (no residual issue).
- **Contract tests**: 98/98 Codex adapter (27 new: handshake ordering, window gates, legacy verdicts + jitter + late verdict, >200-page walk to the true start, seek cursors, revision stability, expansion, tool-only growth, in-place skeleton freshness, stat-move demotion + zero-churn re-anchor, append rebase, compaction rebuild, concurrent/aborted/preempted materialization, cross-session isolation, forced-walk fallback, circuit-breaker tiers, edited-history re-read, last-page-RPC race, materialized-window LRU, cross-epoch collision recovery) + 31/31 app-server client (3 new `ensureReady`) + viewer end-to-end test pinning that revision-moving tool-only growth republishes while projection-identical refreshes early-return.
- **Perf budgets**: `codexWindowedColdStartMs` 300 (measured ~56ms synthetic), `codexLegacyColdStartMs` 2500 (simulated ~1.3s with the late-verdict pattern); existing budgets unchanged and green.
- **Gates**: `test:ci:linux` exit 0 (unit 811, contract+integration 1352, browser, packaging, coverage), lint baseline, architecture guards, behavior contracts + capability audit (`audit.head` at the fix commit).

### DIFF
--- a/.ci/conversation-performance.json
+++ b/.ci/conversation-performance.json
@@ -9,5 +9,7 @@
   "hostSessionSwitchMs": 250,
   "webviewInitialPublicationMs": 500,
   "webviewIncrementalRefreshMs": 250,
-  "webviewWarmFrameRestoreMs": 100
+  "webviewWarmFrameRestoreMs": 100,
+  "codexWindowedColdStartMs": 300,
+  "codexLegacyColdStartMs": 2500
 }
--- a/.skills/developing-provider-conversation-adapters/SKILL.md
+++ b/.skills/developing-provider-conversation-adapters/SKILL.md
@@ -89,6 +89,13 @@ From `spikes/codex-paginated-read` and `spikes/codex-cold-start`:
   recorded on a `limit:100` summary walk seeks the same turn boundary
   with `limit:4, itemsView:"full"` and returns items byte-identical to
   the full walk.
+- Item ids are NOT reliably turn-scoped while a session is live: a
+  response-spanning item (e.g. a reasoning block) can be projected into
+  different turns by fetches taken at different times (observed on a real
+  183MB session). Treat cross-turn id collisions between separately
+  fetched pages as a mixed-epoch signal (invalidate + re-read), never as
+  a protocol violation — circuit-breaking on a transient kills the
+  accelerator.
 - Version gates need the **completed handshake**: `serverInfo.version` /
   `userAgent` parsing only has a value after `initialize` returns, which
   the first `request()` triggers. Any feature gate evaluated before the
--- a/.skills/developing-provider-conversation-adapters/SKILL.md
+++ b/.skills/developing-provider-conversation-adapters/SKILL.md
@@ -73,3 +73,32 @@ the transcript tail:
 
 Read bounded head/tail windows (a single record can exceed 200KB); never
 scan whole transcripts in the listing path.
+
+## Codex App-Server Pagination Facts (0.147, probe-verified)
+
+From `spikes/codex-paginated-read` and `spikes/codex-cold-start`:
+
+- `thread/turns/list` summary view carries the **first** userMessage
+  per turn (projection schema: `thread_turns.first_user_item_id`),
+  **verbatim** — plus a final agentMessage that is UNRELIABLE: omitted
+  for interrupted turns and divergent in ~1/218 real turns. Never derive
+  fingerprints from the summary agent text. Multi-userMessage turns exist
+  in real data (mid-turn steering messages) and collapse to their first
+  message in summary view — never assume one user message per turn.
+- Page cursors are **portable across `limit` and `itemsView`**: a cursor
+  recorded on a `limit:100` summary walk seeks the same turn boundary
+  with `limit:4, itemsView:"full"` and returns items byte-identical to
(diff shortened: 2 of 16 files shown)


hzcheng:
lgtm

PR #231, opened 2026-08-12
fix: serialize Kimi conversation loads per session to keep the cache converged

hzcheng:
## Summary

Fixes a live incident where a healthy, non-empty Kimi session was persistently misreported as "This AI session has no conversation yet." (`conversation-follow / category: empty / provider: kimi` on every click) while a fresh adapter read the same `wire.jsonl` fine.

**Root cause:** warmup, telemetry polls, watch refreshes, and authoritative clicks all share one `KimiConversationIndex` entry per session, and every `load()` committed that entry in place with no ordering. A read that captured the cached `nextOffset` before a racing commit but re-checked it afterwards flipped its continuation decision, re-read a suffix of `wire.jsonl` as if it were the whole file, and wrote the truncated index back with an end-of-file offset. Every later incremental read returned empty, and each click refreshed the poisoned entry's TTL so it never self-healed.

**Fix:**
- Serialize `KimiConversationAdapter.load()` per session id on a rejection-tolerant promise chain, so every read observes a fully committed entry.
- Record `lastReadContinuation` and expose sanitized `getCacheDiagnostics` (source size, cached offset, cached interaction count, continuation, partial) through a new optional adapter method + coordinator passthrough.
- Enrich `conversation-follow` failure diagnostics with session id hashes, warm/fresh provenance, the discarded-empty-warm flag, outline size, source revision, and the cache state above — a genuinely empty session is now distinguishable from a diverged cache in the local log, with tests pinning that no raw identifiers, paths, or content leak.

**Known follow-ups (not in this PR):**
- `claudeAdapter.ts` has the identical unguarded in-place commit pattern and gets the same serialization treatment next.
- Oversized single interactions (>512KB serialized) still throw `tooLarge` and surface as unavailable; deterministic per-interaction content convergence is a separate PR.

**Behaviors:** `CONVERSATION-CACHE-CONVERGENCE-001` (P0), `CONVERSATION-FOLLOW-DIAGNOSTICS-001` (P1). RED evidence: the staggered-wave race test fails on unfixed code within the first round (3/3 runs, truncated `1 !== 4`); recovery guards and diagnostic-shape tests pass only with the fix.

## Skill harvest

updated .skills/developing-provider-conversation-adapters — recorded the incremental-cache concurrency hazard (in-place commits need per-session serialization) and the staggered-wave RED construction, verified against real on-disk Kimi sessions. Recorded as the `Skill-Harvest` trailer of the capability audit commit.

## Verification

- RED before fix: `node --test tests/contract/aiSessions/kimiConversationAdapter.test.js` — race test fails 3/3 with truncated outline; `getCacheDiagnostics` absent; follow-diagnostic shape tests fail.
- GREEN after fix: focused contract (35) + integration (43) pass; race test stable 5/5; full `npm run test:deterministic:run` 413/413.
- `npm run test:behavior-contracts`, `npm run lint`, `npm run test:architecture-guards`, `npm run test:skills`, `git diff --check` all pass.
- Real-data validation with the compiled adapter against the two incident sessions on disk (`409493a4…` live-growing, `70ce2ce2…` 14.5MB): cold reads + staggered concurrent waves + append waves all converge to cold-read truth; cache diagnostics report expected values.

### DIFF
--- a/.skills/developing-provider-conversation-adapters/SKILL.md
+++ b/.skills/developing-provider-conversation-adapters/SKILL.md
@@ -109,3 +109,24 @@ From `spikes/codex-paginated-read` and `spikes/codex-cold-start`:
   ids, and responseStates are all unchanged. Adapter revisions must
   therefore move on **any** provider content change — including
   summary-invisible tool output — or the webview never re-reads the page.
+
+## Incremental Cache Concurrency
+
+Kimi and Claude adapters keep a per-session incremental index
+(`nextOffset` + parsed interactions) and commit it **in place**. Warmup,
+telemetry polls, watch refreshes, and authoritative clicks all call
+`load()` concurrently, so an unguarded check-then-act across the
+continuation hash await lets a racing load flip `continuing`, re-read a
+file suffix as if it were the whole file, and write the truncated index
+back with an end-of-file offset — the session then reports empty
+forever while every later click refreshes the poisoned entry's TTL.
+Serialize `load()` per session id (a trailing promise chain that
+survives rejections) so every read observes a fully committed entry.
+Regression-test with staggered concurrent waves (0..N `setImmediate`
+ticks apart) of mixed `readOutline`/`readSnapshot`/`readTelemetry` after
+each append, then compare against a cold-read truth adapter — the flip
+window is only a few event-loop ticks wide, so unserialized code fails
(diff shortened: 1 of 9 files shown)


hzcheng:
lgtm

PR #232, opened 2026-08-12
fix: serialize Claude conversation loads per session

hzcheng:
## Summary

Follow-up to #231 (merged): `claudeAdapter.ts` had the identical unguarded in-place cache commit pattern fixed for Kimi there — a load capturing the cached `nextOffset` before a racing commit but re-checking it afterwards flips its continuation decision, re-reads a transcript suffix as if it were the whole file, and writes the truncated index back with an end-of-file offset. A healthy Claude session then reports "no conversation yet" forever while clicks refresh the poisoned entry's TTL.

**Fix (mirrors #231):**
- Serialize `ClaudeConversationAdapter.load()` per session id on a rejection-tolerant promise chain.
- Record `lastReadContinuation` and implement the sanitized `getCacheDiagnostics` introspection from #231, so empty Claude follows carry the same diagnosable cache state.
- `CONVERSATION-CACHE-CONVERGENCE-001` and `CONVERSATION-FOLLOW-DIAGNOSTICS-001` gain Claude owners/evidence.

RED evidence: the staggered-wave race test fails on the unfixed adapter (truncated outline in the first rounds); recovery guards pass pre-fix and pin the required append/replace/truncate semantics.

## Skill harvest

no skill change — the incremental-cache concurrency lesson was already recorded into `.skills/developing-provider-conversation-adapters` by #231; this branch adds no new workflow friction. Recorded as the `Skill-Harvest` trailer of the capability audit commit.

## Verification

- RED before fix: Claude race test fails with truncated outline; `getCacheDiagnostics` absent.
- GREEN after fix: focused contract suite 31/31; race test stable 3/3; full `npm run test:deterministic:run` 413/413.
- `npm run test:behavior-contracts`, `npm run lint`, `npm run test:architecture-guards`, `git diff --check` all pass.
- Real-data validation with the compiled adapter against real on-disk Claude sessions (16MB/49 turns, 23MB/215 turns): cold reads + staggered concurrent waves + append waves all converge to cold-read truth; cache diagnostics report expected values.

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -4956,10 +4956,12 @@
     "priority": "P0",
     "status": "automated",
     "owners": [
-      "tests/contract/aiSessions/kimiConversationAdapter.test.js"
+      "tests/contract/aiSessions/kimiConversationAdapter.test.js",
+      "tests/contract/aiSessions/claudeConversationAdapter.test.js"
     ],
     "evidence": [
-      "src/aiSessions/conversation/kimiAdapter.ts"
+      "src/aiSessions/conversation/kimiAdapter.ts",
+      "src/aiSessions/conversation/claudeAdapter.ts"
     ]
   },
   {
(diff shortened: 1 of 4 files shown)


hzcheng:
lgtm

PR #233, opened 2026-08-12
fix: converge oversized conversation turns instead of failing the read

hzcheng:
## Summary

Fixes the second incident from the field: a Kimi session whose conversation could not be opened at all — "Unable to read the AI session conversation. Click the session again to retry." — because one serialized interaction exceeds the 512KB `maxPageBytes` protocol budget.

**Root cause:** per-field caps (64k-grapheme messages, 4k tool call details) never bound an interaction's *aggregate* size. The incident session (14.5MB `wire.jsonl`, 45 turns) carries four turns with 100+ tool calls and 200–300KB of merged thinking each. `buildConversationPage` could only drop whole interactions; with one oversized interaction left it threw `tooLarge`, `readSnapshot` failed, the outline vanished with it, and composition mapped the error to `unavailable` (with a pointless retry).

**Fix (deterministic in-interaction convergence, `model.ts`):**
- Keep the user message and the latest assistant/plan/question endpoints; turns with no semantic endpoint (real orchestrator turns whose content lives in provider events the adapter skips) fall back to the last worklog entry.
- Water-fill byte budgets across preserved messages; truncate only allowlisted content fields (`markdown`/`text`/`detail`/`question`/`label`/`description`/`otherLabel`) with grapheme-safe UTF-8 byte truncation; tool names, diff paths, plan file paths, question sources, and outcomes survive verbatim.
- Insert an explicit omission notice ("Work was omitted to keep this turn within the conversation size limit.") and greedily backfill recent messages that still fit.
- Every interaction block is bounded to `maxPageBytes − 2KB`, so a single-interaction page always fits; the shrink loop now only decides how many interactions fit, and `tooLarge` remains as unreachable defense. Outline construction is untouched and keeps all 45 interactions.

Credits: direction and byte accounting adapted from the uncommitted candidate in `.worktree/kimi-conversation-empty`, simplified (allowlist field truncation instead of reflective blacklisting) and extended (endpoint-less turns, real-shaped fixtures).

**Behavior:** `CONVERSATION-OVERSIZED-TURN-001` (P0). RED evidence: all 8 new tests fail on unfixed code — the real-shaped contract test throws exactly `ConversationError: tooLarge` at `buildConversationPage`. Mutation check: adding `filePath` to the truncation allowlist makes the plan-identity test fail.

## Skill harvest

updated .skills/developing-provider-conversation-adapters — recorded the aggregate-budget gap behind `tooLarge`, empirical fixture calibration (130 capped 4KB details ≈ 504KB < 510KB budget; 160 needed), and endpoint-less complete turns from real data. Recorded as the `Skill-Harvest` trailer of the capability audit commit.

## Verification

- RED before fix: 5 unit + 2 contract + 1 integration tests fail (tooLarge / missing function); real-shaped fixture calibrated against the incident turn's measured shape (128 tool calls, ~255KB thinking, ~278KB tool output).
- GREEN after fix: focused suites pass (unit 15, contract 37, integration 93); full `npm run test:deterministic:run` 817+950+414, zero failures.
- `npm run test:behavior-contracts`, `npm run test:skills`, `npm run test:architecture-guards`, `npm run lint`, `git diff --check` all pass.
- Real-data end-to-end with the compiled adapter on the incident session: snapshot opens (45 interactions, default page 419KB), all 45 anchored pages fit the budget, exactly the 4 oversized turns render bounded with user-first + omission + tail content preserved.

### DIFF
--- a/.skills/developing-provider-conversation-adapters/SKILL.md
+++ b/.skills/developing-provider-conversation-adapters/SKILL.md
@@ -130,3 +130,24 @@ within the first waves while the serialized fix is deterministically
 green. Record `lastReadContinuation` on the entry and expose
 `getCacheDiagnostics` so an empty follow is diagnosable from the local
 log without raw identifiers.
+
+## Page-Budget Convergence and Endpoint-Less Turns
+
+Per-field caps (64k-grapheme messages, 4k tool details) do NOT bound an
+interaction's aggregate size: 100+ tool calls or merged thinking runs sum
+past `maxPageBytes`, and `buildConversationPage` used to have no
+in-interaction lever — it threw `tooLarge` and the whole conversation
+reported unavailable. Converge inside the interaction instead: keep the
+user message and the latest assistant/plan/question endpoints, truncate
+only allowlisted content fields (`markdown`/`text`/`detail`/`question`/
+`label`/`description`/`otherLabel` — never tool names, diff paths, plan
+file paths, or outcomes), insert an explicit omission notice, and
+backfill the tail greedily. Two calibration lessons from real incident
+data: (1) size synthetic oversized fixtures empirically — per-field caps
+shrink naive fixtures below the page budget (130 capped 4KB tool details
+≈ 504KB < 510KB; 160 needed), so measure the built block before trusting
+a RED; (2) real sessions contain complete turns with NO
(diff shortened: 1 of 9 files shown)


hzcheng:
lgtm

PR #234, opened 2026-08-12
feat: Codex config profile selection for AI sessions

hzcheng:
## Summary

Add first-class Codex configuration profile (`-p <name>`) support for AI sessions:

- **Create**: when at least one `<name>.config.toml` overlay exists in the Codex home (`$CODEX_HOME` or `~/.codex`), the new-session flow asks which profile to use — `Base configuration (no profile)` plus every discovered profile, preselecting the last used one (or the `agentPivot.codexDefaultProfile` setting) by ordering it first. Profiles are probed once per CLI installation (`codex resume --help`); CLIs without `-p` support keep the picker hidden with a one-time upgrade hint.
- **Persist**: the effective decision is stored per session as a three-state record (`profile` / explicit `base` / no record = legacy) only after the runtime actually starts, migrated from pendingId to session id at promotion, copied on runtime rebind, and deleted on archive. The store writes atomically (temp file + rename) and warns visibly on save failures.
- **Resume**: always reuses the decision recorded at creation — recorded profile resumes with `-p <name>`, recorded base and legacy sessions resume without one. The default-profile setting never leaks into resume. A recorded profile whose file disappeared prompts to continue with the base configuration or open settings.
- **Visible**: history and active rows show an accessible profile badge (tooltip + aria-label, truncation, `unavailable` marker when the file is gone). The telemetry model chip is unchanged and keeps showing the actually running model from the rollout.

Profile names are basename-validated (Unicode and inner spaces allowed, traversal/separators/leading dashes rejected) and passed as separate argv elements through the existing shell escaping. Kimi and Claude paths are untouched.

## Skill harvest

no skill change — the only process friction hit (architecture-guard fixture registration, safety-check text assertions) is already documented in resilient-webview-mutation-protocols and review-fix-commit-loop; these were compliance gaps, not instruction gaps.

## Verification

- `npm run test-compile` — clean
- `npm run test:deterministic:run` — 825 unit + 978 contract + 413 integration tests pass
- `npm run test:safety:run` — workspace parity / tmux / AI session safety / open projects / skills checks pass (launchOptions import assertion updated for the added reader)
- `npm run test:behavior-contracts` — passes; 7 new `SESSION-CODEX-PROFILE-*` contracts registered and assigned to MAIN-WORKSPACE-SCOPE
- `npm run lint` — 0 errors, warning count unchanged vs base (99)
- `node scripts/run-dashboard-webview-checks.js` — pass
- `git diff --check` — clean

Known trade-offs (documented in README): the profile store is last-writer-wins across concurrent VS Code windows; a shell rc overriding `CODEX_HOME` inside terminals is unsupported; choosing "Use Base Configuration" for a missing profile applies per resume and is not persisted over the recorded decision.


hzcheng:
lgtm

PR #235, opened 2026-08-13
feat: one-click quick AI session creation with provider memory

hzcheng:
## Summary

Add one-click quick AI session creation to the AI SESSIONS header: a split button whose quick side launches a session with the workspace's remembered provider (no pickers), and whose arrow opens a keyboard-operable dropdown with one quick entry per provider plus the full interactive flow.

- `createSessionQuick()` on the creation controller skips every picker, resolves the Codex profile from the last-used/configured default without prompting, defers the multi-root choice to the remembered directory scope, and persists the started provider per workspace on a dedicated `workspaceQuickCreateAiSessionProvider.v1` key (legacy key used only for migration fallback; the list-filter selection can no longer clobber the memory).
- The provider memory write is awaited before the reveal/refresh so the following hydration already sees it; the quick button, its aria label, and a static caption line under the button (`Codex · deepseek` / `Kimi`) all follow the remembered provider, plumbed through hydration as `quickCreateProvider`/`quickCreateProfile`.
- Dropdown contract: shared page-level menu, focus moves in on open, Arrow/Home/End rove, Enter/Space activate, Escape restores focus to the originating arrow, outside clicks close, and the arrow toggles with mirrored `aria-expanded`.
- Fixes found while testing: split-button SCSS was nested inside the legacy `.ai-session-create-button > span:first-child` rule and never matched (restored + raised the expanded CURRENT WINDOW pane floor 250 -> 263 for the caption line); stale `media/` webview mirrors synced; anonymous-navigation privacy check bounded to rendered content.

Behavior contract `AI-SESSION-QUICK-CREATE-001`; capability `MAIN-AI-SESSION-QUICK-CREATE` with all implementation commits assigned.

## Skill harvest

no skill change — every miss this branch exposed (SCSS anchor mangling, unsynced media/ mirrors, missing verification before commit) is already covered by review-fix-commit-loop and resilient-webview-mutation-protocols; the failures were compliance gaps, not instruction gaps.

## Verification

- `npm run test-compile`, `npm run lint` (baseline warnings only), `git diff --check`
- `npm run test:deterministic:run`: 831 unit + 994 contract + 415 integration, 0 failures
- Browser suite: 237 passed (quick-create split button, dropdown keyboard/toggle/outside-click, caption at rest, remembered-provider override, pane floor at 263px)
- `node scripts/run-dashboard-webview-checks.js`, `npm run test:safety:run`, `npm run test:behavior-contracts` — all green
- Mutation check: removing the provider-memory `await` fails the new ordering test as expected
- Rendered screenshots reviewed at 315px and 240px sidebar widths
- Built via `npm run install-local`; installed bytes hash-verified against the VSIX (dist/dashboard.js, media/styles.css, webview scripts)


hzcheng:
lgtm

PR #243, opened 2026-08-16
fix: log silent worktree group provisioning failures

hzcheng:
## Summary

A worktree task whose member provisioning fails **after** the physical worktree was created (finalize / manifest write stage) only persisted an error code into the in-memory member record. Nothing reached the output channel, so post-hoc diagnosis was impossible — this made a real `arch-refactor` task failure undebuggable after the fact.

This change adds host-side logging to the silent paths, with no behavior change:

- `IsolatedSessionController` settles: log any non-success outcome with kind / errorCode / operationId via the previously unwired `onSettled` option.
- `WorktreeGroupCreationController.settleMemberOutcome`: log kind / errorCode / completedSteps / groupId / memberId before the failed state write.
- `runMember`: the settle-outcome persist catch no longer swallows silently; it logs with the outcome summary.
- `recordProvisionedWorktree`: wrap in try/catch and log the original stack before rethrowing (the throw degrades the operation to a retryable partial).
- `DashboardDiagnostics.logError`: tolerate a `undefined` error argument for message-only entries instead of printing a bare `undefined` line.

## Verification

- `npm run test-compile` (after rebase onto `origin/main` @ b67050dd)
- `node --test tests/contract/worktrees/groupCreationController.test.js tests/contract/worktrees/provisioningController.test.js tests/contract/aiSessions/runtimeComposition.test.js` — 56 pass
- `npm run test:behavior-contracts` — pass
- `git diff --check` — clean

## Skill harvest

none — the only process friction was a malformed patch hunk caught immediately by `test-compile`; `review-fix-commit-loop` already covers anchor verification, so no skill change is justified.

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "e670a7b0b7d2d81297e1386eb7b5f809de181ee0",
+        "head": "ecdfa526e6a26c7c154b1d7bb0374219ffa0a36d",
         "ignoredDocumentationCommits": [
             "dd86a325e3db5aa013ffa18a647e67e9fa279c10",
             "0b0e12b4d97ec7d77a8a93076459fdaad2e52070",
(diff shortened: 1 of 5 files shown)


hzcheng:
lgtm

PR #245, opened 2026-08-16
fix: render Codex /goal continuation turns in the conversation viewer

hzcheng:
## Problem

Codex `/goal` sessions were broken in AI Conversation in two ways:

1. The `/goal <objective>` input was never displayed.
2. After a goal run, everything the agent did inside goal-continuation turns disappeared from the conversation and the outline.

Root cause (probe-verified against real rollouts and the 0.147 app-server): `/goal` never produces a `userMessage`. The objective is recorded only as a `thread_goal_updated` event and in the injected `<codex_internal_context source="goal">` user message — which the app-server strips from turn items. Goal-continuation turns therefore arrive from `thread/read` / `thread/turns/list` with **no userMessage at all**, and the adapter dropped every reasoning, tool-call, plan, and agent item in them because no interaction existed to attach them to. One observed session hid a single turn containing ~4 hours of work (229 reasoning / 348 commandExecution / 72 agentMessage / 209 fileChange items).

## Fix

- New `CodexRolloutGoalTurnsReader` incrementally scans the rollout JSONL and maps goal-continuation turn ids to their objective, anchored on the injected goal message's `<objective>` tag under the turn's `task_started` (exact even when a thread replaces its goal mid-session).
- The Codex adapter synthesizes a `/goal <objective>` interaction for those turns — in both the full normalization path and the windowed skeleton projection (same deterministic `<turnId>-goal` id, so anchors survive materialization) — giving the turn's content an interaction to attach to. Unknown user-less turn kinds keep the previous hidden behavior.
- Wired through `composition.ts` per the guarded Codex adapter boundary; the adapter only gains one structural option (`readGoalTurns`).

## Verification

- New contract tests: full-read rendering (messages, timing, outline), conservative hiding without rollout goal data, and windowed skeleton/page parity with the full-read ground truth; plus incremental-reader unit tests. Codex adapter contract file: 106/106 green.
- Real-data check per `developing-provider-conversation-adapters`: compiled adapter + real app-server against a completed-goal session — the `/goal …` outline entry appears and its 342 messages render.
- `npm run test:ci:linux` fully green (lint, behavior contracts, architecture guards, coverage gates).

## Skill harvest

updated: `.skills/developing-provider-conversation-adapters` — recorded the probe-verified 0.147 layout fact that `/goal` continuation turns lose their userMessage in `thread/read` and the objective survives only in the rollout transcript (and `goals_*.sqlite`), so adapters must seed those interactions from a rollout scan. Validated with `tests/unit/tooling/repositorySkills.test.js` and `scripts/run-skill-management-checks.js`; recorded as the `Skill-Harvest` trailer in the capability-audit commit.


hzcheng:
lgtm

PR #246, opened 2026-08-16
feat: reclaim collapsed OPEN tab regions for information density

hzcheng:
## Summary

Collapsed regions on the OPEN tab no longer hold space they cannot use:

- **Collapsed OPEN WINDOWS** shrinks to its title bar and docks to the panel bottom; CURRENT WINDOW is uncapped — in the sidebar fit layout an expanded card fills all freed space, in the wide legacy pane it stays content-sized but uncapped.
- **Collapsed CURRENT WINDOW card** falls back to content size so OPEN WINDOWS packs upward; a dragged separator share is *suspended* (never discarded) while either region is collapsed and restored once both are expanded again.
- **The separator only participates while both regions show content** — it hides when the CURRENT WINDOW card is collapsed or OPEN WINDOWS is collapsed, and leaves the keyboard order while hidden.

Implementation notes:

- The wrapper state class `open-windows-collapsed` is synced from `syncResizer` (runs after every optimistic toggle and authoritative replacement) because webview Chromium older than Chrome 105 cannot match `:has()` for sibling state — same pattern as the existing `current-card-expanded` class.
- Collapse-reclamation CSS is placed after the sidebar fit rules so equal-specificity ties resolve toward reclamation.
- Behavior contract `WEBVIEW-OPEN-TAB-SPLIT-001` re-pinned: the separator now requires an expanded CURRENT WINDOW card; existing drag/restore/floor tests were switched to expanded-card fixtures, and the dropped squeezed-region scroll assertions were redundant with the layout test's scroll-independence coverage.

## Verification

- `tests/browser/dashboardOpenTabSplit.test.js` — 12/12 pass, including new geometry assertions for the four collapse states at 800px (wide), 360px (default sidebar), and 240px (minimum sidebar): collapsed bar bottom-docked within 2px, share suspension releasing a 70% pin back to content size, share restoration on re-expand, no page scroll.
- `npm run test:browser:run` — 309/309 pass (full browser suite, catches resizer-visibility fallout in other tabs).
- `node scripts/run-dashboard-webview-checks.js` — pass.
- `npm run test:behavior-contracts` — pass (catalog title updated for the reclamation behavior).
- `git diff --check` — clean.
- Screenshots captured at 360px/240px for the collapsed states: `/tmp/open-tab-*.png` (bounding-box assertions are the primary rendered-output evidence).

## Skill harvest

updated:.skills/review-fix-commit-loop — this task hit a real workflow trap: `npx gulp` (default task) ends in watch mode under development and never exits, stalling asset rebuilds; pinned `npx gulp --production` as the one-shot rebuild command (the mode CI uses).


hzcheng:
lgtm

PR #247, opened 2026-08-17
In-conversation worktree changes panel (task result + SCM-grouped working changes)

hzcheng:
## Summary

Adds an in-conversation worktree changes view (PRD: `docs/worktree-changes-panel-prd.md`): a Changes pill on the conversation telemetry strip plus a read-only sidebar panel that answers "what did this task change" without leaving the conversation — task result relative to a frozen per-member baseline, plus uncommitted working changes grouped the way Source Control groups them.

### Baseline data contract (P0 prerequisite)

- Every worktree group member captures an immutable `MemberBaseline` (frozen `baseRef^{commit}` SHA + branch/tag/commit source) before any physical side effect; provisioning branches from the frozen SHA, never from a base ref that may advance mid-flight, and verifies HEAD against the baseline after creation
- Corrupt or absent baselines degrade explicitly (`Baseline unavailable` / `History rewritten`) — never guessed, never rendered as zero
- The conversation worktree resolver returns manifest-compatible `WorktreeKey`s and accepts Windows paths

### Changes collection and panel

- Self-collected `git status --porcelain=v1 -z --untracked-files=all` grouped into the four SCM groups (merge/staged/changes/untracked), independent of the user's `git.untrackedChanges` setting; cross-member aggregation with a complete/partial/unavailable tri-state
- Telemetry pill shows bare counts and opens the sidebar view; the sidebar tab row was removed — the four telemetry pills are the view switchers, with a pressed background on the active view
- Member dropdown (repo · branch, single-repo groups degrade to a read-only label), collapsible directory tree of files, click opens the native diff in the current window (index↔worktree / HEAD↔index / direct open / merge entry), Review opens the baseline multi-diff, plus an "Open in Source Control" jump
- Refresh rides Git state-change events with the telemetry cycle as fallback plus a manual refresh; the panel follows session switches, resumes, and mid-session membership changes

### Dogfooding fixes included

- Porcelain XY protocol validation rejected space-bearing codes (`' M'`, `'M '`) — the most common entries — silently swallowing file clicks
- Changes messages were stamped with the activation-time viewer generation and silently dropped after the first rebind, freezing the panel; publish now stamps the current generation, and the webview adopts new generations on session reset with a self-healing refresh pull
- The member dropdown rebuilt its options on every state push, tearing down the open native popup (flicker); rebuilds are now skipped while focused and when the visible signature is unchanged
- Icon sizing joined the shared telemetry selector groups; tooltips no longer double up (custom + native) and wrap long paths
- Viewer load/switch failure paths now emit sanitized diagnostics to the Agent Pivot output log, and resync requests carry the sanitized apply error

### Codex adapter fix (found while dogfooding)

- Verified on a live resumed codex 0.147 thread: resume/compact epochs restart per-item numbering, so item ids legitimately repeat across turns. Item identity is now scoped as `turnId · itemId`; same-turn duplicates still fail closed. Previously every resumed or compacted codex session failed with "unable to read the AI session conversation"

## Test plan

- [x] `npm run test-compile`
- [x] Unit: 1066 pass
- [x] Contract: 1164 pass (incl. real-git worktree provisioner and codex adapter epoch tests)
- [x] Integration: 426 pass (incl. end-to-end viewer + real collector + real git worktree chain)
- [x] Browser: 312 pass (panel rendering, dropdown stability, scrolling, session-switch following, degraded states)
- [x] Rendered-output verification via `scripts/dev/changes-panel-screenshot.js` (icon parity with sibling telemetry buttons, no overflow at default/min widths)
- [x] `npm run test:behavior-contracts`; capability audit current (`MAIN-WORKTREE-CHANGES-PANEL`, codex epoch fix under its existing capability)
- [x] Real-data verification per adapter skill: live resumed codex thread read through the production normalizer

## Skill harvest

Updated `.skills/developing-provider-conversation-adapters`: recorded that codex item ids are only unique within a turn (resume/compact epochs restart numbering), with the turn-scoped keying rule. No other skill changes were justified.


hzcheng:
lgtm

PR #248, opened 2026-08-17
fix: worktree surface resilience — refresh-stable forms, surface retention, chip collapse, conversation header identity

hzcheng:
## Summary

Four annotation-driven fixes for the worktree task surface and the conversation viewer:

- **fix: preserve form focus and scroll through authoritative dashboard refreshes** — the inline worktree group creation form lost input focus (and any in-progress IME composition) whenever an authoritative `ai-sessions-updated` / `open-workspaces-updated` refresh replaced the DOM, and the current-workspace list / window scroll snapped to the top. The form's focus is now captured before the replacement and restored by `reconcileDom`; both scroll regions are carried across the replacement; and the Host skips posting byte-identical `ai-sessions-updated` messages for every refresh reason instead of only `watcher`/`attention`, eliminating no-op replacements entirely.
- **fix: keep the worktree surface when a conversation closes** — closing the conversation viewer hard-coded a switch back to the Chats surface. The focus-origin handler now probes the Worktree surface for the origin session first (the same follow-the-session rule `revealAiSessionInWorkspace` uses) and only falls back to Chats when the session is not there.
- **feat: collapse worktree group chips beyond four and floor the task name width** — multi-root task group rows let repository chips squeeze the task name away. Chips now collapse into an accessible `+N` marker past four (PRD §10), each chip keeps its full-name hover tooltip, and the title keeps a 10ch floor.
- **feat: render project · task · session in the conversation header with click-to-rename** — the viewer identity line gains the worktree task group name resolved from the session's worktree key (active and history sessions; hidden for group-less sessions), and the session name is now a button that asks the Host to run the existing rename flow, with the authoritative reconcile updating the header in place.

## Skill harvest

no skill change — the version-skew strip flow and the media byte-identity requirement are already documented in `resilient-webview-mutation-protocols`; the one slip (media copy not regenerated before the first push attempt) was caught immediately by the existing `webviewState` byte-identity test, so this is a compliance gap, not an instruction gap.

## Verification

- `npm run test-compile` (tsc + dashboard bundle rebuild) — clean
- `node --test tests/unit tests/contract tests/integration` — 2657 passed
- `npm run test:browser:run` — 320 passed (including new coverage: form focus survives both authoritative replacement paths, open-workspaces scroll preservation, chip `+N` collapse, worktree surface retention on conversation close, conversation header identity line and rename intent, rename protocol validation)
- `npm run test:safety:run`, `run-dashboard-webview-checks.js`, `run-conversation-performance-checks.js`, `npm run lint:ci`, `npm run test:behavior-contracts`, `git diff --check` — all green
- Capability audit re-keyed after the rebase onto `main` (`docs: re-key capability audit after rebase`); `audit.head` advanced to the last implementation commit
- Installed locally via `SKIP_NPM_CI=1 npm run install-local` and byte-verified the packaged VSIX against the installed extensions directory (SHA-256 on `dist/dashboard.js`, `media/webviewDashboardBundle.js`, `media/conversationViewerScripts.js`, manifests)


hzcheng:
lgtm

PR #249, opened 2026-08-17
fix: merge streamed Kimi text deltas into one conversation block

hzcheng:
## Summary

A Kimi session's AI Conversation rendered **every one or two characters as its own line** (reported against session `c6113be4`).

Root cause, confirmed against the session's `wire.jsonl`: the Kimi wire protocol streams assistant text as **token-sized `ContentPart` deltas** (this session had a run of 406 one-to-two-character deltas) interleaved with empty `think` keep-alives. The adapter coalesced *thinking* deltas via `pendingThinking` but appended every *text* delta as a separate `assistantMarkdown` entry, so each delta rendered as its own block. The per-delta `normalizeVisibleText` pass also trimmed whitespace-only deltas, silently eating the spaces between words.

Fix (mirrors the existing thinking path):

- Buffer raw text deltas per run (`pendingText`); publish them as one normalized block so whitespace-only deltas survive.
- The buffer persists across incremental loads; a partial run **replaces** its trailing block on the next load instead of appending.
- Empty think parts are streaming keep-alives and no longer split the surrounding text run.
- Any non-text event (tool call, plan, question, approval, turn boundary) closes the run.

## Verification

- New RED→GREEN contract test: token deltas + a space + empty think keep-alives, split across two incremental loads, render as exactly one `**3c 完成**` block.
- Real-session verification: the reported session's delta-storm turn now renders as a single 873-character assistant block.
- `tests/contract/aiSessions` + `tests/unit/aiSessions`: 929/929 pass.
- `npm run test:coverage:ci`: changed-line coverage 96.09% (gate: 80%).

## Skill harvest

none — the investigation followed existing debugging/verification skills; no reusable workflow gap surfaced.

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "0fed732eee67b22fe3e041711a80f16b213fc52b",
+        "head": "3b8064d460afec74027e39d4c8f7e8120e9658af",
         "ignoredDocumentationCommits": [
             "dd86a325e3db5aa013ffa18a647e67e9fa279c10",
             "0b0e12b4d97ec7d77a8a93076459fdaad2e52070",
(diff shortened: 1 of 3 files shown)


hzcheng:
lgtm

PR #250, opened 2026-08-17
fix(worktrees): consume preview token atomically across the baseline window

hzcheng:
## Summary

The worktree-group confirm path validated the preview snapshot, then awaited per-member baseline resolution, and only then deleted the preview snapshot. Two confirms carrying the same token could both pass validation while parked in the baseline-resolution window and provision the same plan twice (two groups with identical branches/paths; the second fails at git level and leaves failed member rows to dismiss). An in-flight confirm could also delete a newer preview snapshot that had replaced its own, making the user's next confirm fail with `preview-stale` despite confirming exactly what they saw.

The fix re-validates snapshot identity at the consumption point so a superseded or already-consumed token fails closed as `preview-stale`, matching the documented single-use semantics. Two regression tests pin the baseline-resolution window and the superseded-preview case.

Found during the architecture program pilot deep dive (documented as finding P5 in `docs/architecture/stage-1b-pilot-rfc.md`); handled as a standalone bug fix per the charter's bug-fix workflow, outside any migration slice.

## Skill harvest

no skill change — the existing fixing-regressions-with-ci and review-fix-commit-loop flows covered this fix; no new lesson justified a skill edit.

## Verification

- RED first: both new tests fail on the unfixed code with the expected defect signatures (`['created','created']` double provision; in-flight confirm wins over the newer preview).
- Focused: `tests/contract/worktrees/groupCreationController.test.js` 26/26; worktrees unit 168/168; worktrees contract 112/112.
- `npm run test:deterministic:run` green (exit 0); `npm run test:behavior-contracts` green; `git diff --check` clean.
- Capability audit commit: assigned to MAIN-WORKTREE-CHANGES-PANEL, with WORKTREE-GROUPS-CREATE-001 added to its behavior set.

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "3b8064d460afec74027e39d4c8f7e8120e9658af",
+        "head": "bb90d8af28fc1f539b638c11ec25c924ac90ed08",
         "ignoredDocumentationCommits": [
             "dd86a325e3db5aa013ffa18a647e67e9fa279c10",
             "0b0e12b4d97ec7d77a8a93076459fdaad2e52070",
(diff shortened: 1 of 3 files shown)


hzcheng:
lgtm

PR #252, opened 2026-08-17
docs: adopt the serial single-worktree program workflow

hzcheng:
## Summary

Owner decision: one job = one worktree = one branch. Every commit for a line of work lands on that worktree's branch, and PRs are cut serially from the same branch as earlier ones merge.

- `AGENTS.md` non-negotiable 1 now states the job-per-worktree rule: no branch switching inside the worktree, no spreading one job across worktrees, no second worktree for the same job. Worktree creation happens once per job. No specific worktree or branch path is baked into repo-level guidance.
- Charter Sections 16 and 20 updated to match: the architecture program runs entirely on the program worktree's branch, with each slice rebased onto the latest `origin/main` before its PR.
- `.skills/protecting-main-with-worktrees` now states the same rule and warns that an open PR absorbs anything pushed to its branch — this skill's per-topic creation example was the remaining guidance source that read as 'new task = new worktree'.

Process/documentation only; no production code or behavior changed.

## Skill harvest

updated .skills/protecting-main-with-worktrees — encoded the owner-confirmed job-per-worktree rule; the skill's per-topic worktree creation example was the remaining misleading guidance source.

## Verification

- `npm run test:behavior-contracts` green; `git diff --check` clean; `tests/unit/tooling/repositorySkills.test.js` 10/10 green (skill pins intact).
- Capability audit: commits assigned to MAIN-REGRESSION-CI-CURRENCY (guidance-layer currency convention).

### DIFF
--- a/.skills/protecting-main-with-worktrees/SKILL.md
+++ b/.skills/protecting-main-with-worktrees/SKILL.md
@@ -26,9 +26,10 @@ job.
    - identify the intended repository remote and base branch from the user request, local tracking branch, or remote default
 
 2. If the user asks for a worktree under the current project, place it under
-   `.worktree/<topic>` — worktree creation happens once per line of work.
-   - Do not add `.worktree/` to tracked `.gitignore` unless the user explicitly wants a repo change.
-   - Prefer local ignore: `printf '.worktree/\n' >> .git/info/exclude` if needed.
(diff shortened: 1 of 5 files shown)


hzcheng:
lgtm

PR #253, opened 2026-08-17
fix: allow removing worktree records whose directory is already gone

hzcheng:
## Summary

A worktree deleted externally (`rm -rf`, or pruned elsewhere) left its group member **impossible to remove** from the CURRENT WINDOW task view: the snapshot resolves the entry as `prunable`/`missing`, `resolveTarget` rejects non-normal health, and the blocker check answered `worktree-not-removable` at deletion admission — so both the journaled group-member deletion and the interactive removal were stuck forever (reported against the `harness-registry` and `fix-preview-token-consumption` tasks).

A physically absent directory is a *certain* observation, not an uncertain one — it must not fail-closed:

- `getBlocker` answers `null` when the path is missing (still fail-closed for detached repositories / truncated discovery where the path exists).
- `removeVerified` short-circuits to the verified absence.
- The interactive `remove()` still confirms, then succeeds.
- All three paths prune the stale git administrative entry (best-effort; skipped when the repository itself is out of view).
- The confirm dialog label falls back to the stale snapshot branch ref, then the path basename, instead of dereferencing a null target.

## Verification

- New RED→GREEN contract tests with a real git repository: externally deleted directory marked `prunable` in the snapshot (and a fully-pruned variant) no longer block `getRemovalBlocker`, `removeVerified` reports `removed`, interactive removal confirms and succeeds, and the stale `git worktree` metadata is pruned.
- Worktrees contract + unit suites: 282/282 pass.
- `npm run test:coverage:ci`: changed-line coverage 88.14% (gate: 80%).

## Skill harvest

none — straightforward RED→GREEN fix against an existing behavior (`WORKTREE-MANAGED-CLEANUP-001`); no workflow friction worth codifying.

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "17ab4dde576090e944e6ea67c3db8133929a3ffb",
+        "head": "e55c43231f82f1e8cd61bf0c99f95f098e38773e",
         "ignoredDocumentationCommits": [
             "dd86a325e3db5aa013ffa18a647e67e9fa279c10",
             "0b0e12b4d97ec7d77a8a93076459fdaad2e52070",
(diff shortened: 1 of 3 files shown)


hzcheng:
lgtm

PR #255, opened 2026-08-18
fix: open task review as one multi-diff with valid vscode.changes triples

hzcheng:
## Summary

The worktree changes panel's **Review** button silently opened only the first changed file instead of the native multi-diff.

Root cause: `vscode.changes` validates its resource list as **[label, original, modified] triples** (see `extHostApiCommands.ts` → `_workbench.changes` → `editorCommands.ts`: each entry decomposes as `[label, original, modified]`, and the validator rejects any entry whose length is not exactly 3). The call passed two-element `[original, modified]` pairs, argument validation rejected the command, and the silent `catch` fell back to opening a single-file `vscode.diff` — with no log anywhere.

Fixes:

- Pass `[fileUri, baselineContentUri, worktreeUri]` triples so the native changes editor opens **every** changed file at once (baseline → worktree).
- Deleted files render an empty modified side instead of a missing working-tree document.
- The capability fallback now matches the PRD (§5.3): a per-file quick pick that opens the chosen diff — and the `vscode.changes` rejection is logged via the dashboard output channel instead of being swallowed.

## Verification

- New unit tests with a real git fixture: triple arity + per-side URIs (baseline ref on the original side, `~empty~` sentinel for deletions), no spurious `vscode.diff` on the primary path; fallback path asserts the quick pick, the chosen file's diff, and the logged error.
- `tests/unit/aiSessions/conversationChangesController.test.js` + `tests/integration/dashboard/conversationViewer.test.js`: 104/104 pass.
- `npm run test:coverage:ci`: changed-line coverage 94.44% (gate: 80%).
- `npm run lint`: 99 warnings — identical to the base (no new warnings).

## Skill harvest

none — the fix followed the existing RED→GREEN and verification skills; no reusable workflow gap.

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "a76abb00ff83dd937011ac4afcd1f60d88dc473e",
+        "head": "b76f4ba375337b0674dc846d119d5ce7f5c46ba8",
         "ignoredDocumentationCommits": [
             "dd86a325e3db5aa013ffa18a647e67e9fa279c10",
             "0b0e12b4d97ec7d77a8a93076459fdaad2e52070",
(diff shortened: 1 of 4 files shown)


hzcheng:
lgtm

PR #256, opened 2026-08-18
feat(architecture): module boundary guards and cycle debt ratchet (Harness v0 PR 2)

hzcheng:
## Summary

Stage 2 PR-2 of the architecture harness program, on top of the merged registry kernel (#254).

- **`scripts/architecture/buildDependencyGraph.js`**: full local import graph over the 366 classified production files — static imports, re-exports, type-only imports, `require()`, bare side-effect imports, and inline `import()` type references; fail-closed on unresolvable local specifiers. Bounded cycle reporting via direct 2-cycles plus Tarjan SCC clusters (enumerating all simple cycles explodes exponentially on this graph — measured 3,283 with type edges).
- **`scripts/architecture/checkModuleBoundaries.js`**: R-BOUNDARY (every resolved cross-module edge must be declared in `mayDependOn`), R-ENTRYPOINT (edges must land on declared public entrypoints — whole-module in the coarse phase, narrowed per wave), R-CYCLE (exact-fingerprint ratchet: new cycle debt fails, stale baseline entries fail), and waiver-ledger coherence (bijection with the baseline, milestone retirement, no wildcards).
- **`scripts/architecture/updateArchitectureDebtBaseline.js`**: reviewed rule-by-rule baseline generation; refused in CI.
- **First debt ledger**: `.ci/architecture-debt-baseline.json` records the measured 16 two-cycles and the 13-module cyclic cluster; `docs/testing/architecture-waivers.json` pairs them with four waivers carrying owner, reason, and retirement milestones (WAVE-W3, WAVE-W4, PILOT-STRICT, WAVE-FINAL).
- Registry corrections the guards surfaced: public entrypoints widened to whole-module surfaces (coarse phase), and two type-only edges to `MOD-SKILL-MANAGEMENT` declared.
- **`tests/unit/architecture/moduleBoundaries.test.js`**: 11 tests — positive fixtures and controlled negative mutations for every rule (undeclared edge, narrowed entrypoint, new cycle debt, stale baseline entry, waiver incoherence).

No production source file changed; observable behavior untouched.

## Skill harvest

no skill change — the guard-design lessons here (bounded SCC reporting instead of simple-cycle enumeration; entrypoints whole-module during the coarse phase) are already recorded in the versioned RFC and will feed the planned guardrails skill per charter Section 14.

## Verification

- `npm run test:architecture-policy`: 30/30 green; classification and boundary checks pass on the real repository.
- `npm run test:behavior-contracts` green after the audit commit; `git diff --check` clean.
[...]

PR #257, opened 2026-08-18
feat(architecture): invariant catalog and single-writer enforcement (Harness v0 PR 3)

hzcheng:
## Summary

Stage 2 PR-3 of the architecture harness program, on top of the merged boundary guards (#256).

- **`docs/testing/architecture-invariants.json`**: the eight approved pilot invariants normalized from the RFC draft — structured authority/writers/enforcement, every module/capability/path reference cross-validated.
- **`scripts/architecture/checkSingleWriters.js`**: catalog schema validation plus the mechanical single-writer check: a write-method call on a declared state family's store outside its writer set fails. The check requires a store reference alongside the method call, so same-named methods on unrelated services (todo groups) are not false positives — verified by a dedicated mutation test.
- **Measured pilot writer sets declared**: the worktree manifest family (7 writer files — the multi-writer reality behind findings P1/P2, ratcheted so migration can only shrink it) and the provisioning store family (2 writer files).
- **`tests/unit/architecture/singleWriters.test.js`**: 10 tests covering catalog schema mutations and the writer scan.
- Lane `test:architecture-policy` now runs all three checkers (closed-world, module boundaries, single-writer).

No production source file changed; observable behavior untouched.

## Skill harvest

no skill change — no new workflow lesson beyond what the harness itself now encodes.

## Verification

- `npm run test:architecture-policy`: 40/40 green; all three checkers pass on the real repository.
- `npm run test:behavior-contracts` green after the audit commit; `npm run test-compile` clean; `git diff --check` clean.
- Architecture scripts line coverage: 87.7%.

### DIFF
--- a/docs/testing/architecture-invariants.json
+++ b/docs/testing/architecture-invariants.json
@@ -1,0 +1,207 @@
+{
+    "version": 1,
+    "title": "Architecture invariant catalog (pilot scope)",
+    "notes": [
+        "Normalized from docs/architecture/drafts/pilot-invariants.draft.json (approved at the Stage 1B gate).",
+        "Every reference is cross-validated: module ids exist in architecture-modules.json, capabilities exist in main-capability-coverage.json, and every path exists in the repository.",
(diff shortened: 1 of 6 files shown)


hzcheng:
lgtm

PR #260, opened 2026-08-18
test(worktrees): close the pilot safety-net gaps (Stage 3)

hzcheng:
## Summary

Stage 3 of the architecture program: the pilot safety net, closing the three failure-matrix blind spots recorded in the approved Stage 1B RFC.

1. **Real-git group creation end to end** (`tests/contract/worktrees/groupCreationRealGit.test.js`): preview and confirm run against two real on-disk repositories with the real provisioner — no fakes in the side-effect path. Asserts both worktrees materialize on branches cut at the frozen baseline SHA and the manifest converges to ready members carrying real worktree keys and baselines.
2. **Confirm handler chain**: the five worktree-group form handlers move **verbatim** from the composition root into `src/dashboard/worktreeGroupFormHandlers.ts` (pure move, no logic change), making the accepted→settled seam testable. `tests/integration/dashboard/worktreeGroupFormHandlers.test.js` locks the chain: parse fail-closed, every accepted request owes exactly one terminal settlement even when the controller throws, and add-repo target forgery is dropped.
3. **Replay asymmetry locked**: the confirm family has no replay cache; a duplicate confirm now has a message-level characterization test showing it reaches the controller and settles failed as `preview-stale` via the single-use preview token (fixed in #250).

Adds behavior contract `WORKTREE-GROUPS-CREATE-HANDLER-001`; extends `WORKTREE-PROVISIONING-GIT-001` owners. Capability audit assigns the change to `MAIN-WORKTREE-CHANGES-PANEL` and links the new behavior.

## Skill harvest

no skill change — the extraction followed the existing review loop; nothing new to encode.

## Verification

- New tests: real-git e2e 1/1, handler chain 8/8; focused suites: worktrees contract 115/115, dashboard integration 10/10.
- `npm run test:architecture-policy` green (the extracted handler file auto-classified into MOD-DASHBOARD-SHELL/application — closed-world working as designed); architecture guards green.
- New source file measures 89% statement coverage from the handler tests.
- `npm run test:behavior-contracts` green after the audit commit; `git diff --check` clean.

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -2178,7 +2178,8 @@
     "priority": "P0",
     "status": "automated",
     "owners": [
-      "tests/contract/worktrees/gitWorktreeProvisioner.test.js"
+      "tests/contract/worktrees/gitWorktreeProvisioner.test.js",
+      "tests/contract/worktrees/groupCreationRealGit.test.js"
     ],
     "evidence": [
       "src/worktrees/gitWorktreeProvisioner.ts"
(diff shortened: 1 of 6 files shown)


hzcheng:
lgtm

PR #261, opened 2026-08-18
feat: merge untracked into Changes and compress single-child folder chains

hzcheng:
## Summary

Two working-changes panel density fixes from UX feedback:

1. **Untracked merges into Changes.** The working-changes layer rendered four sections (Merge → Staged → Changes → Untracked); the untracked section only repeated what the per-row `U` badge already says. Sections now render Merge → Staged → Changes with untracked rows inside Changes, badges intact. Row click semantics stay exact: `item.group` still flows to the open-file intent.
2. **Single-child directory chains compress into one row** (`a/b/c`), aligned with Source Control's tree. A chain level qualifies only while it holds one child directory and no direct files; deep Java paths (`com/xhs/reddb/service/impl/Foo.java`) now take two folder rows instead of five indented ones. Collapse state keys anchor to the chain's final directory.

PRD §行为明细 updated to match.

## Verification

- RED→GREEN in `tests/browser/conversationViewer.test.js` (real Chromium): group headers are now `['Staged Changes', 'Changes']`, the untracked row keeps its `U` badge inside the merged section, `src/auth` renders as one folder row, and the deep-path fixture renders exactly `['com/xhs/reddb', 'service/impl']` with the file row at two indent levels; compressed rows keep working collapse toggles.
- Full `conversationViewer.test.js`: 114/114 pass; dashboard webview checks pass.
- `npm run test:coverage:ci` pass; `npm run test:behavior-contracts` pass; `git diff --check` clean.

## Skill harvest

none — small UX iteration with the existing RED→GREEN loop; no reusable workflow gap.

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "8d27256aaefe7d1eee5c021cf8058da9e2a58c8c",
+        "head": "33dd08d563b7d9e1e04352c4869af6f10fad9b2e",
         "ignoredDocumentationCommits": [
             "dd86a325e3db5aa013ffa18a647e67e9fa279c10",
             "0b0e12b4d97ec7d77a8a93076459fdaad2e52070",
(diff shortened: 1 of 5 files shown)


hzcheng:
lgtm

PR #270, opened 2026-08-18
fix(architecture): enforce exact-one role and revert W1 strict to migrating (review R1)

hzcheng:
## Summary

W1 review Blocker 1 + the ledger revert. The closed-world loader used first-match-wins for roles, so 106 production files matched multiple roles while red line 1 claims exactly one.

- **`loadArchitecturePolicy.js`**: roles now form an exact partition — specific role globs must be disjoint (an overlap fails with file, module, both roles, and the matching patterns); a role with `include: ["**"]` is the remainder role, allowed only in last position, catching the module's unclaimed files. Empty role include lists are now rejected (they previously passed), and an invalid module can no longer crash classification mid-run.
- **Registry**: remainder roles applied across modules; `ntc.ts` removed from its double assignment; the shell's empty domain role removed; the 'first matching role wins' note is gone.
- **`checkProgramLedger.js`**: a strict module now fails while the closed-world classification has any error — red line 1 is a mechanical strict precondition.
- **W1 ledger reverted to `migrating`** per the review; strict re-acceptance follows the repair list (R1–R10).
- New controlled mutations: role overlap fails, misplaced remainder role fails, remainder catches only unclaimed files, strict-with-classification-errors fails.

No production source file changed; harness and policy data only.

## Skill harvest

no skill change — the failure mode (partition claims without partition enforcement) is exactly what the harness now encodes.

## Verification

- Architecture policy lane 69/69 green; classification passes with 371 files / 16 modules and zero role overlaps.
- Full coverage gate locally: baseline passed, changed-line coverage 97.78% against origin/main.
- `npm run test:behavior-contracts` green after the audit commit; `git diff --check` clean.

### DIFF
--- a/docs/architecture/changes/ARCH-CHANGE-003.md
+++ b/docs/architecture/changes/ARCH-CHANGE-003.md
@@ -1,0 +1,51 @@
+# ARCH-CHANGE-003 — Role semantics: exact partition with a remainder role
+
+Date: 2026-08-18
+Status: approved by repository owner (review repair R1)
+Module: all (registry-wide semantics)
+
+## Problem and evidence
+
+W1 review Blocker 1: the closed-world loader used first-match-wins for
+roles — 106 production files matched multiple roles while red line 1 claims
+exactly one per file. The registry note even documented "the first matching
+role wins", contradicting the red line.
+
+## Old rule → new rule
+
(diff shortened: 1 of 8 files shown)


hzcheng:
lgtm

PR #271, opened 2026-08-18
feat(architecture): protect the harness surface in the change gate (review R2)

hzcheng:
## Summary

W1 review Blocker 2: the anti-self-amendment gate only protected the policy JSONs — guard implementations, guard tests, and CI wiring could be weakened inside a product-only PR.

- **`reportArchitectureDiff.js`** gains the protected harness surface (`scripts/architecture/**`, `tests/unit/architecture/**`, `.github/workflows/**`, `run-architecture-guards.js`, `ciContracts.js`, `package.json`) and computes the weakening delta: deleted harness files, removed guard ids, removed lane/workflow invocations, shrunk mutation tests.
- **`checkArchitectureChange.js`**: any harness weakening classifies as relaxing and requires an ARCH-CHANGE record; a guard-logic change with intact wiring is tightening (the kill mechanism is the guard's own mutation tests).
- Controlled mutations for all four review scenarios (guard file deleted, guard id removed, lane invocation removed, mutation tests shrunk) plus the intact-wiring tightening case and a fake-git end-to-end delta test.

## Skill harvest

no skill change.

## Verification

- Architecture unit 17/17 green; policy lane 69/69 green; the gate self-classifies this PR.
- Full coverage gate locally: baseline passed, changed-line coverage 87.83% against origin/main.
- `npm run test:behavior-contracts` green after the audit commits; `git diff --check` clean.

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "e628ec8365f5312e5d547feef7b08606228e9d99",
+        "head": "2b7bdcc70d234bed812e572897b323623790eb9b",
         "ignoredDocumentationCommits": [
             "dd86a325e3db5aa013ffa18a647e67e9fa279c10",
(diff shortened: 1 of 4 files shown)


hzcheng:
lgtm

PR #288, opened 2026-08-19
feat(architecture): type-resolve the single-writer guard (round-2 Important 1)

hzcheng:
## Summary

Fixes round-2 review Important 1: the single-writer guard's name-matching scan could be laundered by barrel imports (the file never mentions the store's lowercase basename) and structural-typing injections (an anonymously typed parameter).

The guard now resolves the receiver's declaring class through the TypeScript checker:

- write-method calls on a store-class-typed receiver fail outside the declared writers — through barrels, aliases, element access, and destructuring;
- a store-class value passed as a call argument into a non-writer file is a provision bypass (structural injections need a provision site to do harm);
- only class instances carry write capability — plain interfaces declared in the store file (member records) are data and do not false-positive (this refinement was caught by the guard firing on real code during development);
- violation messages keep the historical substrings so the guard mutation parity lane's base suites still kill.

Residual documented in the guard header: a writer deliberately handing the store to a helper in another file is writer discipline, covered by review.

## Change impact declaration

```change-impact-declaration
{
  "headSha": "38d6e57fb33d6c246c215e0af747320606ce21d6",
  "capabilities": [
    "MAIN-ARCHITECTURE-HARNESS"
  ],
  "modules": [],
  "invariants": [],
  "policyDelta": "tightening",
  "newFiles": []
}
```

## Skill harvest

none — no skill change justified; the slice followed the established review-fix-commit loop.

## Verification

- `node --test tests/unit/architecture/singleWriters.test.js` — 25/25 ✅ (new controlled mutations: barrel import, structural injection provision site, read-not-flagged)
- `node scripts/run-guard-mutation-parity.js` ✅ — base suites green against the new guard (message-compatible strengthening)
- `npm run test:architecture-policy` ✅ / real-repo single-writer check ✅
- `npm run test:coverage:ci` ✅ — changed-line 100% (135/135), real exit code
- `git diff --check` ✅

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "0b255233c0d74e05f662c66a52c7e6725e602357",
+        "head": "fc8b08177c37bd47ad7ff2175900a2ad4b6bb573",
         "ignoredDocumentationCommits": [
             "dd86a325e3db5aa013ffa18a647e67e9fa279c10",
             "0b0e12b4d97ec7d77a8a93076459fdaad2e52070",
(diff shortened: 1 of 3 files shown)


hzcheng:
approve 38d6e57fb33d6c246c215e0af747320606ce21d6

PR #289, opened 2026-08-19
feat(architecture): make Architecture Change records exact and scoped (round-2 Blocker 3)

hzcheng:
## Summary

Fixes round-2 review Blocker 3: Architecture Change records were reusable wildcard authorizations — `modules` was never consulted during matching, subset semantics let a declared superset authorize more than the owner reviewed, and records outlived their delta in the base forever.

- **Exact per-dimension equality** replaces subset semantics; a declared superset now fails (the R3 test pinning subset coverage is inverted).
- **Module scope enforced**: the record's `modules` must cover every module the change actually touches — touched production modules, delta map keys, both ends of every `fileMoves` entry, ledger-regression subjects, and changed invariants' modules.
- **File-moving re-partitions** declare per-file `fileMoves: [{ path, from, to }]`; the gate computes actual moves by classifying changed files against the base registry.
- **Invariant semantic changes** declare per-invariant field lists with before/after sha256 fingerprints (`invariantChanges: [{ id, fields, before, after }]`); removed invariants declare the removed record's fingerprint. New `scripts/architecture/describeArchitectureChange.js` prints the record-ready delta for authors.
- **Reuse analysis** (for the reviewer): with exact equality, re-consuming a record means re-realizing the identical delta — once merged, the same diff no longer occurs, so a record cannot authorize a *different* change. Records 004/005 (already consumed) are non-candidates under the new schema.
- Charter 8.9 documents the exact-equality + scope + fingerprint contract.

## Change impact declaration

```change-impact-declaration
{
  "headSha": "d14d7653f56ce3884f0d78831ee4c5bd5fdaec55",
  "capabilities": [
    "MAIN-ARCHITECTURE-HARNESS"
  ],
  "modules": [],
  "invariants": [],
  "policyDelta": "tightening",
  "newFiles": []
}
```

## Skill harvest

none — no skill change justified; the slice followed the established review-fix-commit loop.

## Verification

- `node --test 'tests/unit/architecture/**/*.test.js'` — 134/134 ✅ (inverted subset test, exact-equality dimension matrix, module-scope mutation, fingerprint mismatch, fileMoves, describer self-diff)
- `npm run test:architecture-policy` ✅ (gate self-classification: tightening)
- `npm run test:deterministic:run` ✅ — unit 1237 + contract 1170 + integration 435
- `npm run test:coverage:ci` ✅ — changed-line 89.95%, real exit code
- `git diff --check` ✅


hzcheng:
Closing temporarily: this change needs the guardSemantics enabler + an ARCH-CHANGE record (guardSemantics) in base for the guard-mutation-parity lane to pass legitimately. Will be re-proposed after those land.

PR #291, opened 2026-08-19
docs: record ARCH-CHANGE-007 (guard semantics evolution authorization)

hzcheng:
## Summary

ARCH-CHANGE-007 record: owner-reviewed authorization for the guard-semantics contract change needed by the record-precision repair (T3 / round-2 Blocker 3).

The record-precision repair changes the verification contract from subset to exact equality for Architecture Change records. The guard-mutation-parity lane needs this base-landed record to exempt the intentional guard-contract change.

This is a docs-only record — no code changes.

## Change impact declaration

```change-impact-declaration
{
  "headSha": "110f0e45cd3c9cf2badd3289a75ad35bef77e73f",
  "capabilities": [],
  "modules": [],
  "invariants": [],
  "policyDelta": "product-only",
  "baselineWaiverDelta": "zero",
  "newFiles": []
}
```

## Skill harvest

none — docs-only record, no code or skill change.

## Verification

- `npm run test-compile` ✅
- `npm run test:architecture-policy` → product-only ✅
- `npm run test:behavior-contracts` ✅
- `git diff --check` ✅

PR #292, opened 2026-08-19
feat(architecture): make Architecture Change records exact and scoped (round-2 Blocker 3)

hzcheng:
## Summary

Round-2 review Blocker 3: Architecture Change records were reusable wildcard authorizations. This makes them exact and scoped.

**Problem**: records used subset semantics, modules was never consulted during matching, and records outlived their delta in the base forever.

**Fix**:
- Exact per-dimension equality replaces subset semantics; a declared superset now fails
- Module scope: the record's modules must cover every module actually touched
- File moves: re-partitions declare per-file fileMoves [{path, from, to}]
- Invariant fingerprints: semantic changes declare sha256 fingerprints via describeArchitectureChange.js
- Same-PR rejection: a record added in the same PR as its consumption is rejected

ARCH-CHANGE-007 (landed in #291) authorizes the guard-semantics change.

## Change impact declaration

```change-impact-declaration
{
  "headSha": "4b2ceebe6e7514c388fc64cfce1bfb9ffd37404a",
  "capabilities": [
    "MAIN-ARCHITECTURE-HARNESS"
  ],
  "modules": [],
  "invariants": [],
  "policyDelta": "tightening",
  "newFiles": []
}
```

## Skill harvest

none — no skill change justified; the fix followed the established review-fix-commit loop.

## Verification

- 40/40 architecture change tests pass
- npm run test:architecture-policy -> tightening
- node scripts/run-guard-mutation-parity.js -> guardSemantics exempted
- npm run test:behavior-contracts pass
- npm run test-compile pass
- git diff --check pass


hzcheng:
approve 4b2ceebe6e7514c388fc64cfce1bfb9ffd37404a

PR #297, opened 2026-08-19
feat(architecture): wire approve-architecture consumption (transition to PR 3/6)

hzcheng:
## Summary

Transition PR for the Harness Simplification program (between PR 2/6 and PR 3/6). The decision record (docs/architecture/harness-simplification-decision.md) replaced Architecture Change record machine authorization with owner architecture approval (`approve-architecture <full-head-sha>`), but the merged gates never consumed it. This wires the consumption path so PR 3/6 (which deletes the record machinery and therefore cannot be authorized by a record) can be evaluated and merged.

Root causes fixed:

- **merge-approval gate**: evaluated only `approve <sha>` and base records — a relaxing change that may not add records (ADR: no new records during the window) could never be authorized. The gate now detects the owner's `approve-architecture <full-head-sha>` comment and threads the verdict into the declaration evaluation.
- **quality lane**: the architecture change gate demanded `approve-architecture <sha>` but nothing consumed the comment. The classifier now accepts an `architectureApproved` verdict; the verify workflow detects the owner comment and exports `ARCHITECTURE_APPROVED` (PR-lane feedback only — the merge gate re-verifies from the default branch).
- **trusted kernel**: could not materialize the PR head (the `pull_request_target` checkout fetches only the base ref, so both `git checkout <sha>` and `git diff base...head` failed) and never received the approval flag. The workflow now fetches the PR head first; the kernel looks up the owner's approval comment via the GitHub API (fail-closed).
- **architecture-approval job**: diffed the base branch against itself (`git diff origin/main HEAD` where HEAD is the base checkout) and always passed vacuously. It now diffs the base against the fetched PR head SHA.

## Skill harvest

Updated `.skills/fixing-regressions-with-ci`: when a CI check fails, first identify which side of the PR supplies the check's code — `pull_request` checks run PR-head files, `pull_request_target` checks (merge-approval gate, trusted kernel) run default-branch files, so a PR-head edit cannot change a default-branch check until it merges. This is the lesson from PR #296's repeated CI failures.

## Known red check (bootstrap)

`trusted-kernel` will fail on this PR: the check runs the **default branch's** kernel, which cannot yet fetch the PR head (this PR is the fix). After this PR merges, the kernel on main becomes functional and evaluates PR #296 with the approval consumption above.

## Change impact declaration

```change-impact-declaration
{
  "headSha": "6cc7fca077b4929411d6dbaede79e8e560346ece",
  "capabilities": [
    "MAIN-ARCHITECTURE-HARNESS"
  ],
[...]

PR #302, opened 2026-08-20
feat(worktrees): narrow the manifest store behind a facade (PR 5/6)

hzcheng:
## Summary

PR 5/6 of the Harness Simplification program. Structural single-writer enforcement for the worktree group manifest store, replacing the type-resolved method scan for this state family.

- **The store class leaves the entrypoint**: cross-module consumers receive a capability-free `WorktreeGroupManifestStoreHandle` plus narrow `WorktreeGroupManifestReader` / `WorktreeGroupManifestWriter` views. A write call outside the 5-method writer view is a compile error; importing the store file from outside the module is an entrypoint violation.
- **Intra-module writers** (8 files) take the handle and unwrap it via the module-internal `worktreeGroupManifestStoreOf`, which is never re-exported.
- **dashboard.ts** (composition root) holds only the views; its write surface shrinks from 25 methods to the 5 it legitimately needs. The unlisted `renameGroup` enumeration gap disappears with the scan.
- **checkSingleWriters**: a `stateFamily` with `writerFacade: true` is enforced by a structural import rule on the dependency graph — only declared writers and the module entrypoint may import the store file. Families sharing one store must agree on the flag. Non-facade families (provisioningStore) keep the type-resolved scan during migration.
- Invariant catalog: both groupManifest families declare the facade.

Mutation-sensitivity proof: a probe file importing the store from `src/todos/` fails the guard (`single-writer: src/todos/__sneaky_probe.ts imports facade store ...`); removing it restores green.

## Skill harvest

no skill change — the facade pattern followed existing codebase conventions (injected closures, handle unwrap); the CI workflow lessons from this program were already harvested in #297/#299.

## Owner approvals

```
approve-architecture c7033e76275cfd4aa48e267a49a61f86cb347b39
```

```
approve c7033e76275cfd4aa48e267a49a61f86cb347b39
```

## Verification

- `npm run test-compile`
- `node --test tests/unit/architecture/** tests/unit/tooling/**` — 546 pass (3 new facade edge-rule cases)
- `node --test tests/unit/worktrees/**` — 191 pass; full unit/contract/integration suites pass
- `npm run test:behavior-contracts` / `npm run lint` — pass
- Default-branch gate evaluation replayed against the exact head: classification relaxing, zero errors with approval
- `node scripts/run-trusted-kernel.js` — passes with approval, fails closed without

### DIFF
--- a/docs/testing/architecture-invariants.json
+++ b/docs/testing/architecture-invariants.json
@@ -1,341 +1,343 @@
 {
-    "version": 1,
-    "title": "Architecture invariant catalog (pilot scope)",
-    "notes": [
-        "Normalized from docs/architecture/drafts/pilot-invariants.draft.json (approved at the Stage 1B gate).",
-        "Every reference is cross-validated: module ids exist in architecture-modules.json, capabilities exist in main-capability-coverage.json, and every path exists in the repository.",
-        "Records with a stateFamily block are mechanically checked by scripts/architecture/checkSingleWriters.js: a write-method call outside the declared writers fails. The writer set may only shrink during migration."
-    ],
-    "invariants": [
-        {
-            "id": "ARCH-WORKTREE-PREVIEW-TOKEN-001",
-            "module": "MOD-WORKTREE-LIFECYCLE",
-            "productCapabilities": [
-                "MAIN-WORKTREE-CHANGES-PANEL"
-            ],
-            "priority": "P0",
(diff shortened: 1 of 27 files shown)


hzcheng:
approve-architecture c7033e76275cfd4aa48e267a49a61f86cb347b39
approve c7033e76275cfd4aa48e267a49a61f86cb347b39

PR #305, opened 2026-08-20
fix(worktrees): narrow the member lifecycle to its own write family (W1 authority)

hzcheng:
## Summary

Two single-writer families share the worktree group manifest store. The member authority (`WorktreeMemberLifecycle`) unwrapped its handle to the **full** store, so it could reach the manifest-structure family's 20 write methods — the cross-family hole the writer union left behind.

Fix in code structure, not process:

- The store gains a `WorktreeGroupMemberWriter` view (exactly the member family's four declared methods: `transitionMember`, `updateMember`, `setPrimaryMember`, `removeMember`).
- `WorktreeMemberLifecycle` now holds that view. Calling another family's method is a compile error, not a scanner finding.

Mutation-sensitivity proof: adding `this.store.deleteGroup(...)` inside `memberLifecycle.ts` fails compilation with TS2339; the probe was removed after the check.

## Skill harvest

no skill change — the fix reuses the facade view pattern from PR 5/6.

## Owner approvals

```
approve-architecture 25393e066279a6fdd2672698a8d73938a6500584
```

```
approve 25393e066279a6fdd2672698a8d73938a6500584
```

## Verification

- `npm run test-compile` — clean (the narrowing compiles; probe mutation fails as expected)
- `node --test tests/unit/worktrees/**` — 191 pass
- `npm run test:architecture-policy` / `npm run test:behavior-contracts` / `npm run lint` — pass

### DIFF
--- a/docs/testing/main-capability-coverage.json
+++ b/docs/testing/main-capability-coverage.json
@@ -2,7 +2,7 @@
     "version": 1,
     "audit": {
         "base": "2b34c653119bdf480f2af0330ee3809b51441807",
-        "head": "e6fa3a6941c3f8a6704adeb844bb35fafb691caf",
+        "head": "fa834b9d7f1a7da76995f6fe3cb5fb8486ca7397",
         "ignoredDocumentationCommits": [
             "dd86a325e3db5aa013ffa18a647e67e9fa279c10",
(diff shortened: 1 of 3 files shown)


hzcheng:
approve 25393e066279a6fdd2672698a8d73938a6500584

PR #307, opened 2026-08-20
feat: switch the conversation bottom rails to previous and next window

hzcheng:
## Summary

The AI Conversation bottom corner rails rarely paid off as session switchers — per-state session cycling now lives on the status dots, and the Previous/Next Active Session commands remain in the Command Palette — while moving between open windows is the everyday action. The rails now switch focus between open WINDOWS.

- One shared identity-sorted ring of every registered open window: each window anchors on its own navigation identity and steps to its neighbour, so the cycle continues naturally as focus moves (A → B → C → A) with no shared cursor or handoff channel.
- Switching reuses the open-workspaces bridge navigate command (focuses the existing window for the workspace, including SSH/WSL/Dev Container targets); empty registrations and untitled multi-root workspaces are excluded.
- With no other window open, the rails show 'No other open windows to switch to.' instead of switching.
- New behavior contract OPEN-WINDOW-CYCLE-RAILS-001 covers the ring logic, the viewer message dispatch, and the rendered rails.

## Owner approvals

```
approve-architecture 8dd0fc43ba6766488bbff23cc7e5ebb1dd4cf367
```

```
approve 8dd0fc43ba6766488bbff23cc7e5ebb1dd4cf367
```

## Skill harvest

no skill change — the workflow followed existing skills (webview mutation protocols, review-fix-commit-loop) without new lessons worth recording.

## Verification

- `npm run test-compile`, `npm run lint`, `git diff --check`
- Focused suites after the final rebase: unit (windowCycleSwitch ring logic, protocol validation, architecture guards), integration dispatch, browser rails tests
- Full browser suite (322) including the updated rails tests and the version-skew fixture chain
- `npm run test:behavior-contracts` after the capability-audit commit
- `SKIP_NPM_CI=1 npm run install-local` — installed and byte-verified

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ All notable changes to the "Agent Pivot" extension will be documented in this fi
 
 ### Changed
 
+- The AI Conversation bottom corner rails now switch between open WINDOWS
+  instead of adjacent Sessions: Previous/Next Window walk one shared,
+  identity-sorted ring of every open window — each window anchors on its own
+  identity, so the cycle continues naturally as focus moves. With no other
+  window open the rails explain themselves instead of switching. The
+  Previous/Next Active Session commands remain available in the Command
+  Palette, and per-state Session cycling lives on the status dots.
 - The AI Conversation status indicators moved from the header into the bottom
   session-navigation row — alongside Previous/Next Active Session — and are
   now clickable badges that count this window's Sessions by lifecycle —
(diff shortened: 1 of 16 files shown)


hzcheng:
approve 083e2c0d5c37e19e80537d41659a14d55d5ab821

hzcheng:
approve 8dd0fc43ba6766488bbff23cc7e5ebb1dd4cf367

PR #306, opened 2026-08-20
feat(architecture): publish the impact report to the job summary, never as a PR comment

hzcheng:
## Summary

The standing impact-report comment (introduced in #300) generated one notification email per PR for the owner. The report is a review aid, not a conversation — the merge-approval gate now writes it to the workflow **job summary** ($GITHUB_STEP_SUMMARY) and the job log instead of the PR comment thread.

- No more per-PR email from the gate.
- The gate's token drops `issues: write` / `pull-requests: write` (read-only comment access is all approval detection needs) — a nice permission tightening after #303's incident fix.
- The report is still regenerated from the exact head on every gate run; find it under the gate job's "Summary" page.

## Skill harvest

no skill change — owner-feedback-driven UX fix, no reusable workflow lesson beyond the recorded ones.

## Owner approvals

Copy each line into a separate PR comment below (both bind the exact head SHA and expire when the head moves):

```
approve-architecture 745e97a22d044861e00e4e7f6e789109f1d1d25a
```

```
approve 745e97a22d044861e00e4e7f6e789109f1d1d25a
```

## Verification

- `node --test tests/unit/tooling/mergeApprovalGate.test.js` — 5/5 pass (wiring now asserts no PR-comment posting + job-summary publishing)
- `npm run test:behavior-contracts` / `npm run lint` — pass

PR #308, opened 2026-08-20
fix(ci): gate crash — remove the leftover REPORT_MARKER reference

hzcheng:
## Summary

The merge-approval gate crashes on every PR right now: #306 removed the `REPORT_MARKER` constant but left a reference in `renderImpactReport`, so the gate dies with `ReferenceError: REPORT_MARKER is not defined` before posting any status (blocking #307 and everything else).

Pattern-matching wiring tests could not catch it — the gate's new code only executes from the default branch **after** the merge. Fixes:

- `renderImpactReport` no longer emits the removed marker line.
- The gate's `main()` is guarded by `require.main` so the module can be required in tests.
- The wiring test now **calls** `renderImpactReport` — a pattern-only assertion shipped this exact bug once already.
- Skill harvest records the post-merge GREEN blind spot for default-branch-only code paths.

Verified locally: `renderImpactReport` invoked directly renders the report; before this fix it throws the ReferenceError.

## Skill harvest

updated .skills/fixing-regressions-with-ci — post-merge GREEN blind spot: when the changed code path only executes from the default branch after merge, verify by directly invoking the changed entry point locally.

## Owner approvals

```
approve-architecture 449d66444fa7278526c56d55457dbca0004071ef
```

```
approve 449d66444fa7278526c56d55457dbca0004071ef
```

## Verification

- `node --test tests/unit/tooling/mergeApprovalGate.test.js` — pass (now calls the renderer)
- `node --test tests/unit/tooling/repositorySkills.test.js` — pass
- `npm run test:behavior-contracts` / `npm run lint` — pass
- Direct invocation smoke: renderer output verified

### DIFF
--- a/.skills/fixing-regressions-with-ci/SKILL.md
+++ b/.skills/fixing-regressions-with-ci/SKILL.md
@@ -25,6 +25,10 @@ Turn every confirmed regression into a CI-owned behavior before changing product
      check's behavior until it merges. Confirm against the failed run's log
      line provenance; when the fix must land on the default branch first,
      sequence a dedicated PR instead of iterating on the blocked one.
+   - The same blind spot applies to GREEN: when the changed code path only
+     executes from the default branch after the merge, CI on the PR cannot
+     exercise it — verify by directly invoking the changed entry point
+     locally (unit-level call, not only pattern assertions on the source).
 2. **Own the behavior**
    - Read `docs/testing/README.md`.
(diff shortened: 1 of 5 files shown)


hzcheng:
approve-architecture 449d66444fa7278526c56d55457dbca0004071ef
approve 449d66444fa7278526c56d55457dbca0004071ef

PR #314, opened 2026-08-20
refactor(architecture): break the attention cycle trio (MOD-AI-SESSION-ATTENTION wave)

hzcheng:
## Summary

Stage 5 wave close-out for **MOD-AI-SESSION-ATTENTION**. The module was named by three direct 2-cycles (ARCH-WAIVER-001: provider, runtime, workspace). Investigation showed each cycle closed through a small shared piece sitting in the wrong home — so this wave applies one direction rule: *the attention pipeline may depend on the session core, never the reverse.*

The three cuts:

1. **Attention session-key codecs → kernel** (`src/attentionSessionKeys.ts`). `getAttentionRuntimeSessionKey` / `getLogicalAttentionSessionKey` are pure identity codecs that runtime settlement and workspace projections were pulling out of the attention module. `attentionProject.ts` re-exports them (existing readers and contract tests keep their path); the four production consumers now import the kernel directly. New characterization tests pin the byte-stable encodings.
2. **`runtimeOwnership.ts` → kernel** (from `src/workspaces/`). `hasWorkspaceRuntimeContinuity` is shared workspace-runtime infrastructure consumed by four modules; the attention event capability no longer edges into MOD-WORKSPACE-IDENTITY.
3. **`statusCapability.ts` re-owned by MOD-AI-SESSION-ATTENTION** (registry-only). The capability drives the attention dot and constructs the attention evaluation queue — its home was mis-classified, which manufactured the PROVIDER→ATTENTION edge. The runtime lifecycle settlement primitives (`settleAiSessionRuntimeLifecycles`, `runAiSessionRuntimeLifecycleTask`) move from the attention controller into `runtimeSettlementCapability.ts`, their only consumer.

Measured outcome (harness-recomputed):

- All three ATTENTION 2-cycles dissolve; baseline 9 → **6** fingerprints; ARCH-WAIVER-001 keeps only the two non-attention pairs (provider↔runtime, provider↔workspace — those belong to the session-core waves).
- The module's remaining outbound edges are one-directional (→ provider/runtime/kernel/notify), all declared.
- SCC membership is unchanged (10) — the triangle through provider↔workspace keeps ATTENTION in the cluster; per the waiver policy that is cluster-level debt blocking whole-program acceptance, not strict mode.
- MOD-AI-SESSION-ATTENTION flips `legacy → strict` (4th strict module of 16).

## Skill harvest

`.vscodeignore` + packaging pin list now documented as release-artifact anchors in review-fix-commit-loop (lesson from this PR's first CI red).

## Owner approvals

Copy the line below into a PR comment (binds the exact head SHA and expires when the head moves):

```
approve 25d2f6a983a132eb07544f98d25f2581e5b98608
```

## Verification

- `npm run test-compile` — pass
- `node --test tests/unit/**` — 1172 pass (3 new codec characterizations), 0 fail
- `node --test tests/contract/**` — 1182 pass, 0 fail
- `npm run test:safety:run` — pass (the settlement-functions anchor moved with them)
- `npm run test:architecture-policy` / `test:behavior-contracts` / `lint:ci` / webview checks — pass
- `npm run test:release-packaging` — pass (after adding the moved kernel files to the shipping whitelist)
- `npm run test:release-packaging` — pass (after adding the moved kernel files to the shipping whitelist)
- `git diff --check` — clean


hzcheng:
approve 25d2f6a983a132eb07544f98d25f2581e5b98608

PR #316, opened 2026-08-20
refactor(architecture): inject the open-workspaces message builder (MOD-OPEN-WORKSPACE wave)

hzcheng:
## Summary

Stage 5 wave close-out for **MOD-OPEN-WORKSPACE**. The `SHELL ↔ OPEN-WORKSPACE` 2-cycle closed through exactly one value edge: the open-workspaces dashboard controller imported `buildOpenWorkspacesUpdatedMessage` from the composition top.

The fix follows the module's own idiom instead of moving files: the controller already receives ~20 collaborators through its options object, so the message builder becomes one more injected option, supplied by the composition root (`dashboard.ts`). The remaining shell edge is type-only; no file moves, no message-shape changes.

Measured outcome (harness-recomputed):

- 2-cycle `SHELL ↔ OPEN-WORKSPACE` dissolved; baseline 6 → **5** fingerprints; ARCH-WAIVER-002 sheds it; waiver bijection holds.
- SCC membership unchanged (10) — the module stays in the cluster through the provider triangle (OPEN→PROVIDER→SHELL), which belongs to the session-core waves; per policy that is cluster-level debt, not a strict blocker.
- Public entrypoints narrow from `src/openWorkspaces/**` to the 12 consumed files; `focusBridgeChannel` becomes module-internal.
- Ledger: MOD-OPEN-WORKSPACE `legacy → strict` (6th strict module of 16).

Anchors updated in the same commit: three test files and two safety scripts that construct the controller now pass the real builder (the contract tests load it through their vscode-fake loaders since the builder's module graph touches `vscode`).

## Skill harvest

none — but a self-inflicted note: two local red rounds came from me (a same-file parallel edit that silently lost hunks, and a stale graph read taken against that broken state). Both were caught by the suites before push; no repo change needed beyond discipline.

## Owner approvals

Copy the line below into a PR comment (binds the exact head SHA and expires when the head moves):

```
approve 793469b921ceee53cb36d21392ae1ab8ada87992
```

## Verification

- `npm run test-compile` — pass
- `node --test tests/unit/**` 1172 / `tests/contract/**` 1182 / `tests/browser/**` 322 — all pass
- `npm run test:safety:run` / `run-dashboard-webview-checks.js` — pass (parity and open-project safety scripts construct the controller and now inject the builder)
- `npm run test:architecture-policy` / `test:behavior-contracts` / `lint:ci` — pass
- `git diff --check` — clean


hzcheng:
approve 793469b921ceee53cb36d21392ae1ab8ada87992

PR #331, opened 2026-08-22
feat(open-tab): switch the OPEN tab to the persistent WINDOWS switcher and remove the split (PR-B)

hzcheng:
## Summary

PR-B of the OPEN tab redesign (`docs/open-tab-window-switcher-prd.md`), stacked on PR-A (#329, merged): the production DOM now switches atomically to the persistent `WINDOWS N` single-line switcher, and the legacy CURRENT WINDOW / OPEN WINDOWS groups plus the draggable split are removed.

- `open-workspaces-updated` moves to v4 (explicit `windowRowCount` / `currentWindowRowCount` / `navigationWindowRowCount` / `currentDetailCount`; the ack `open-workspaces-rendered` moves to v3 with `hasWindowSwitcher`). Note: this is the host↔webview message version, unrelated to the bridge wire protocol (`OpenWorkspaceAggregateV4`).
- Window rows route clicks through `[data-open-window-row]` before the generic project path: non-current rows drive the navigation settlement protocol from PR-A (pending → focused / inline error + retry), the current row is inert (`aria-disabled`, `aria-current`, `Current window:` accessible name), pin/menu/retry stay row-local; the shared ⋯ menu offers Focus Window / Pin / Save Workspace with full keyboard support.
- The transitional current-detail card stays in a headless `.open-current-workspace-group` shell so the `ai-sessions-updated` channel keeps its replacement unit; the expanded-card fit layout is preserved without the resizer.
- Bridge degradation: non-ready status renders in the fixed switcher slot, the current row pins on top, other rows are disabled and dimmed, and their menu Focus item is hidden (no click queueing).
- Empty windows own a current row (`WINDOWS 1`), per the PRD empty-state clause.
- Split removal: scripts, bundle inputs, manifest entry, `.vscodeignore`, styles (surgical: the transition fit layout is preserved), and every checker anchor (`run-dashboard-webview-checks`, `run-ai-session-safety-checks` including the stale `263` literal trap noted in review, `run-open-project-safety-checks`, `run-workspace-parity-checks`, `run-ai-session-tmux-checks`, `run-skill-management-checks`, `run-release-packaging-checks`, `run-performance-architecture-baseline-checks`, `run-architecture-guards`).
[...]

PR #329, opened 2026-08-21
feat(open-tab): window-switcher navigation protocol, v4 schema, and row renderer (PR-A, additive)

hzcheng:
## Summary

First slice (PR-A, additive-only, no production behavior change) of the OPEN tab redesign per `docs/open-tab-window-switcher-prd.md` (committed on this branch). The persistent `WINDOWS N` single-row switcher replaces the CURRENT WINDOW / OPEN WINDOWS split; this PR lands the parts that carry no user-visible change:

- **Versioned window-navigation request/settlement protocol**: `open-window-navigation-request` / `open-window-navigation-result` (v1) with `requestId + cardId` correlation. Host settles every request exactly once (`focused` / `stale-target` / `untitled-workspace` / `failed` / `malformed-request`); the webview pending manager ignores stale/duplicate receipts, supersedes consecutive clicks, times out into an inline error with retry, and replays pending/error row state after authoritative DOM replacements. Registered in the router but not wired to row clicks until PR-B.
- **`open-workspaces-updated` v4 schema** (defined, not yet posted): explicit `windowRowCount` / `currentWindowRowCount` / `navigationWindowRowCount` / `currentDetailCount` replace the ambiguous `currentWorkspaceCount` ahead of the transition form (current window as both a switcher row and the headless-shell detail card). Note: this is the host↔webview message version, unrelated to the bridge wire protocol (`OpenWorkspaceAggregateV4`).
- **Window-row view model + renderer + styles** (not yet rendered in production): single-line slots, three-state current-row model (current/pending/error), `list`/`listitem` aria model with Tab-reachable buttons, disambiguation via deterministic shortest-unique suffix, responsive width matrix via viewport media queries (no container queries — the extension still supports old Chromium).
- Two new behavior contracts: `OPEN-WINDOW-SWITCHER-UI-001`, `OPEN-WINDOW-NAVIGATION-SETTLEMENT-001`.

PR-B (atomic production DOM switch + split removal + contract migration) follows on this branch after PR-A merges.

## Verification

- `npm run test-compile`, `npm run lint`, `git diff --check` — green
- `tests/unit` 1213 pass, `tests/browser` 329 pass, `tests/integration` 438 pass, `tests/contract` 1182 pass
- `npm run test:behavior-contracts`, `npm run test:safety:run`, `npm run test:architecture-policy` — green
- `node scripts/run-dashboard-webview-checks.js` — green
[...]

PR #332, opened 2026-08-22
fix(open-tab): close the review loop on switcher empty-window, geometry, and focus

hzcheng:
## Summary

Post-merge review follow-ups on the M1 window switcher (#329 / #331). All three P1 (merge-blocking) and three P2 findings are fixed, each with regression coverage.

**P1 fixes**

- **Empty-window `ai-sessions-updated` failed DOM consistency**: the message declared `currentWorkspaceCount: 1` for the zero-root placeholder card while the renderer filters it out, so every watcher update tripped `invalid-ai-session-workspace-update` into a full refresh. The channel now counts only the renderable current card, and the shared `getRenderedCurrentWorkspaceNavigationIdentity` returns `null` for unrendered placeholders so the presentation workspace match agrees with the DOM. New empty-window incremental-update browser test + host-side builder assertions.
- **Bridge status slot displaced rows**: the slot was an empty `<div>` in ready state but gained a margin-carrying `<p>` in degraded states (measured 50px group growth). The slot now has a fixed 24px height in every state with normalized single-line content (ellipsis + full-text tooltips), per the PRD fixed-slot/zero-displacement rule. New geometry-invariance browser test across ready → connecting → unavailable → update-required → ready.
- **Authoritative replacement dropped focus**: only the pin button was captured/restored. Focus is now captured and restored by `{cardId, controlKind}` across the row primary, pin, more, and retry controls, with the navigation reconcile replayed first so a re-unhidden retry stays focusable. `closeMenu()` no longer yanks focus back to the trigger on blur/outside clicks — only Escape (keyboard dismissal) returns focus, matching the PRD keyboard contract.

**P2 fixes**

- **Empty window offered a pin the host always rejects** (placeholder ID fails the protocol's card-ID shape): the view model carries `canPin: false`, the row renders no pin button, and the ⋯ menu hides its Pin item for that row.
- **Late success receipt could not clear the timeout error**: the navigation error state now keeps its `requestId`, and a late `focused` receipt matching it clears the error (a newer request deletes the entry, so a match proves no supersede).
- **Compat/animation budget**: the switcher list reserves its scrollbar lane via `@supports not (scrollbar-gutter: stable) { overflow-y: scroll }` for the declared old-Chromium range, and the pin reorder flash is 150ms as documented (was 450ms in both the animation and the class-removal timer).

**Anchors updated in the same commits**: `ARCH-AI-SESSION-PRESENTATION-TRANSACTION-001` guard + mutation fixtures (rendered-identity implementation line), `ARCH-WEBVIEW-MANIFEST-001` manifest (new consumer of `__agentPivotOpenWindowNavigation`).

## Verification

- `npm run test:ci:linux` green end to end: compile, brand, behavior contracts, lint, coverage (changed lines 100%), browser (incl. 6 new switcher tests), safety, dashboard webview checks, architecture policy/guards/baseline, release notes, release packaging.
- Focused: `tests/browser/dashboardWindowSwitcher.test.js` 16/16; integration dashboard 438/438; contract openProjects 93/93.

## Skill harvest

Updated `.skills/resilient-webview-mutation-protocols`: cross-script `window.__agentPivot*` reads must be declared as consumers in `docs/testing/architecture-webview-manifest.json` (learned from an `ARCH-WEBVIEW-MANIFEST-001` failure this branch introduced and fixed). Validated via `tests/unit/tooling/repositorySkills.test.js`. A second failure (stale `ARCH-AI-SESSION-PRESENTATION-TRANSACTION-001` source anchor after editing the anchored line) is already covered by the anchored-reader rule in `review-fix-commit-loop` — compliance gap, no skill change.

## Owner approvals

```
approve 6ec903c0ffac7fa5ff5f2ccbcf662210e4c3bcaf
```


hzcheng:
approve 6ec903c0ffac7fa5ff5f2ccbcf662210e4c3bcaf

PR #334, opened 2026-08-22
docs: define the Machine Agent product direction

hzcheng:
## Summary

- define Agent Pivot's long-term direction as an IDE-independent, local-first platform built around Machine Agents, an optional Control Plane, and shared task/workflow identities
- specify Machine Agent v1 as the narrow near-term slice for reliable background Attention delivery on Remote SSH and managed tmux sessions
- mark the unimplemented `notifyd` design as historical so future implementation follows the durable Outbox, health, supervision, and rollout contracts in the new specification

## Key decisions captured

- keep the Machine Agent in this repository initially, but build and run it as an independent zero-`vscode` product artifact
- distinguish Task, Session, Run, WorkflowDefinition, and WorkflowRun rather than binding one task to one provider session or workflow
- retain foreground notification mode while making background protection explicit, capability-aware, and opt-in
- require a real Remote SSH process-lifetime spike and disconnect acceptance matrix before implementation planning claims background reliability
- persist per-sink delivery work before sending and record success only after an accepted transport response

## Verification

- `npm run test-compile`
- Markdown parsing and local-link resolution for all three changed documents
- `git diff --check origin/main...HEAD`
- complete merge-base-to-head document review; no unresolved Critical or Important findings

## Skill harvest

No skill change — the only workflow friction was an initial relative lookup of worktree-local skill mirrors; the provided absolute skill locators and existing loading rules already cover the correct behavior, so this was a compliance error rather than a reusable instruction gap.

## Owner approvals

```text
approve 48769c001813bafc8bb5645d90f880ebcc8f14c8
```

### DIFF
--- a/docs/product/agent-pivot-platform-vision.md
+++ b/docs/product/agent-pivot-platform-vision.md
@@ -1,0 +1,420 @@
+# Agent Pivot Platform Vision
+
+> Status: product direction draft
+> Date: 2026-08-22
+> Horizon: long-term product direction, not an implementation commitment
+
+## Purpose Of This Document
+
+Agent Pivot today is a VS Code command surface for switching, monitoring, and
+resuming Codex, Claude, and Kimi sessions. This document describes the larger
+product that the current extension can grow into without making VS Code, one
+AI provider, or one machine the permanent system boundary.
+
+The document is intentionally stable and directional. It defines the product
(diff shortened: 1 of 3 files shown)


hzcheng:
approve 48769c001813bafc8bb5645d90f880ebcc8f14c8

PR #336, opened 2026-08-22
fix: count and review untracked files in worktree task result (+ v2 panel PRD/plan)

hzcheng:
## Summary

Two commits on top of `origin/main`:

1. **docs**: Worktree Changes Panel v2 PRD (merged into `worktree-changes-panel-prd.md` as Part II, five reviewed decision rounds) + implementation plan (`worktree-changes-panel-v2-plan.md` v1.2). Covers the four owner-reported pain points: truncated branch name, dropdown-only repo switching, no collapse/expand-all, and no commit history view.
2. **fix (PR-0 of the plan)**: `taskFileCount` and `openTaskResultReview` only read the tracked diff (`git diff --name-only <baseline>`), so untracked files appeared in Working changes but were missing from the Task result count and the Review multi-diff — breaking the PRD §4.3 promise that Task result ⊃ Working changes. Now both use tracked diff ∪ untracked (`git ls-files --others --exclude-standard`), with untracked files rendered against an `EMPTY_REF` original side.

## Test plan

- RED → GREEN: new contract test (`changesCollector` real-repo fixture: committed + modified + untracked + gitignored) and unit test (`openTaskResultReview` triples include the untracked file with an empty baseline side) failed before, pass after.
- Focused suites: unit/worktrees + unit/services + contract/worktrees 338/338; conversationChangesController 9/9; integration conversationViewer 99/99; browser conversationViewer 115/115.
- `npm run test-compile`, `npm run lint` (touched files clean), `npm run test:behavior-contracts`, `git diff --check` all green.

## Skill harvest

none — no process friction worth harvesting (one self-caught `node --test` directory-invocation slip, fixed immediately; no skill gap).

## Owner approvals

```
approve 2674e2b01fcfa94bb5b6ae36f861d64f768b906f
```

PR #343, opened 2026-08-22
feat(worktrees): conversation changes panel v2 — tracking state, files tree, and review hardening

hzcheng:
## Summary

Worktree Changes Panel v2 — the full feature stack (12 commits on current `origin/main`), superseding #336 (closed; its two commits are included here as rebased equivalents). Implements the v2 PRD/plan from `docs/`:

- **PR-0 fix**: task result counts and the Review multi-diff now include untracked files (tracked diff ∪ `git ls-files --others`), keeping Task result ⊃ Working changes.
- **Tracking-branch state**: members surface tracking info as a three-state union (`tracked` with frozen-SHA ahead/behind, `none` as a stated fact, `unknown` as a failed query) — unknown is never rendered as zero; HEAD is collected even without an upstream so invalidation signatures stay complete.
- **Two-row worktree header**: repo title + branch/tracking rows with ‹ › cycling and a transient pending cursor, so rapid consecutive activations never re-select the same repo.
- **Focusable tooltip overlay**: all truncated text (repo title, branch, task summary, tracking) is keyboard-focusable with a viewport-pinned overlay; native `title` attributes migrated to `data-tooltip`.
- **Collapsible change groups**: per-folder and collapse/expand-all with live DOM-derived fold state; focus is restored to the group header only when the focused row is hidden, never stolen from toolbar buttons.
- **Files tree keyboard model**: roving tabindex, ↑/↓/←/→/Home/End/Enter/Space, and focus recovery across refreshes and collapse-all.
- **First-open width recommendation**: new installs open the panel at 320px (192/240/320 screenshot tiers); user resizes permanently win, and existing users without the flag are untouched.
- **Review-round hardening** (three parallel reviews + two final integration reviews, all findings closed):
  - session-switch/terminal states clear stale panel content; tracking renders independently of task-file collection
  - unknown counts render as `?`, never `0`; tracked∪untracked review fails closed instead of opening partial results
  - directory trees and member contexts use `Map` (prototype-key paths like `constructor/` cannot corrupt the tree)
  - host controller: activation epoch rejects stale async resolutions; `clear()` now resets the changes controller so disposal kills the Git watcher and in-flight collection
  - protocol: changes state dual-publishes v2 + legacy v1 so adjacent old webviews keep working; all five changes action intents are bound to subscription generation + session identity, so intents stranded by a session switch fail closed instead of acting on the new session

## Test plan

- `npm run test-compile` ✔
- Browser suite 142/142 ✔ (incl. sha256-pinned adjacent-generation script fixture, updated for the merged strip chain)
- Integration suite 101/101 ✔ (incl. new: disposal resets the changes controller/watcher; stranded changes actions dropped across a session switch)
- Unit: changesController + viewerProtocol 20/20 ✔ (incl. new: every changes action validates generation/session binding)
- `node scripts/run-dashboard-webview-checks.js` ✔, `npm run test:behavior-contracts` ✔, `npm run test:architecture-policy` ✔, `npm run lint` ✔ (touched files clean)
- `npm run test:coverage:ci` ✔ — baseline green, changed-line coverage 99.26% vs `origin/main`
- `git diff --check` ✔; `src/webview` ↔ `media` byte-identity ✔
- Final read-only integration review of the complete merge-base→HEAD diff: **READY TO PUBLISH**, zero Critical/Important/Minor findings

## Skill harvest

- **Rebase across a sha256-pinned script fixture**: when `main` and the branch both evolve the pinned previous-script strip chain, neither pinned hash survives the merge — recompute the reconstruction hash at HEAD (eval the chain offline, verify with the fixture test) instead of picking a side. Candidate addition to `review-fix-commit-loop` or the version-skew note in `resilient-webview-mutation-protocols`; deferring the edit until this flow repeats once more.
- **Generated assets in rebases**: `media/*.css` conflicts resolve by regenerating from the merged `.scss` (`npx gulp buildStyles`), never by picking a side — already covered by existing skills; no change.
- Otherwise no process friction worth harvesting.

## Owner approvals

```
approve 16c903d42c8c12e6945f5b79deefe2bd78a1275c
```


hzcheng:
approve 16c903d42c8c12e6945f5b79deefe2bd78a1275c

PR #356, opened 2026-08-24
fix(windows): prioritize attention count

hzcheng:
## Summary

- Render attention counts before running-session counts in Windows rows.
- Keep the behavior catalog and unit/browser regression coverage aligned with the slot order.

## Validation

- `npm run test-compile`
- `node --test tests/unit/webview/windowSwitcherContent.test.js`
- `node --test tests/browser/dashboardWindowSwitcher.test.js`
- `npm run test:dashboard:run`
- `npm run install-local` (packaged and byte-verified both extensions)

## Skill harvest

No skill change: the task-local packaging timing observation is already covered by the existing verified-artifact and byte-verification guidance; no reusable instruction gap was found.

## Owner approvals

```text
approve a06c33bcbfcb1cf210425d5f438effd1c47fa628
```

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -2934,7 +2934,7 @@
   {
     "id": "OPEN-WINDOW-SWITCHER-UI-001",
     "domain": "webview",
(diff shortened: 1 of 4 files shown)


hzcheng:
approve a06c33bcbfcb1cf210425d5f438effd1c47fa628

PR #372, opened 2026-08-24
style(open-tab): refine the Windows rail into a quiet edge-to-edge navigation list

hzcheng:
## Summary
- Redesign the OPEN tab Windows rail: rows are now edge-to-edge and span the exact same width as the Worktree Group cards below (the inset card chrome and the reserved scrollbar gutter are gone; both regions share one overflow model, so they stay aligned with or without scrollbars).
- Resting window rows stay quiet — transparent surface, hairline status tick, muted icon chip; emphasis is reserved for hover (soft surface lift, hairline ring, brightened tick) and the current window (soft selection surface with a top sheen, hairline ring, and a single focus-color accent bar; semibold name). State changes animate with 120ms ease-out transitions, disabled under reduced-motion.
- Worktree groups get the layered widget surface, a full-width toolbar divider, a visible keyboard-focus outline, and a 32px header rhythm matching the window rows; toolbar action hit areas stay 24px under the taller rhythm.
- Browser coverage: quiet-resting/emphasis contract for window rows, hover visibility without widget theme tokens, and a cross-region assertion that window rows and worktree groups share the same left edge, width, and 32px vertical rhythm.

## Verification
- `npm run test-compile`
- `node --test tests/browser/dashboardWindowSwitcher.test.js` (23 pass)
- `node --test tests/browser/worktreeGrouping.test.js tests/browser/worktreeGroups.test.js` (49 pass)
- `node --test tests/browser/activeSessionCardInteraction.test.js` (51 pass)
- `npm run lint` (pre-existing warnings only)
- `git diff --check`
- Rendered geometry verified at 320px and 420px via a production-DOM Playwright harness: window row and worktree group both measure x=10, width=300 (320px viewport)
- `npx gulp --production` one-shot build is byte-identical to the committed `media/styles.css`

## Skill harvest
no skill change — both friction points in this round were compliance gaps, not instruction gaps: the `npx gulp` default watch-mode hang is already documented in `.skills/review-fix-commit-loop` and `.skills/installing-vscode-extensions-locally` (one-shot `npx gulp --production`), and the 24px hit-area regression came from running only a subset of the browser suites, which the fresh-verification rules in `.skills/review-fix-commit-loop` already cover.

## Owner approvals
```text
approve f59c5a719b5f303c6983ad7beaa5e9cdf9d9ad72
```


hzcheng:
approve f59c5a719b5f303c6983ad7beaa5e9cdf9d9ad72

PR #373, opened 2026-08-24
Perf: render recent conversation messages first

hzcheng:
## Summary
- Render the recent portion of a long latest conversation first, then deliver the complete transcript after acknowledgement.
- Keep progressive delivery resilient across recovery, authority rollover, and superseding publications.
- Announce deferred content through the existing trusted live status region.

## Testing
- `npm run test-compile`
- `node --test tests/integration/dashboard/conversationViewer.test.js`
- `node scripts/run-dashboard-webview-checks.js`
- `git diff --check origin/main --`

## Skill harvest
No skill change: the review-driven protocol fixes followed the existing resilient Webview mutation and review/commit-loop guidance; no reusable instruction gap was found.

## Owner approvals
```text
approve 67d75ba98385e90c458e062bfd98bc08143209ec
```

PR #374, opened 2026-08-25
perf: backfill long conversation history in ack-paced chunks

hzcheng:
## Summary

After a progressively rendered partial page (recent messages first, PR #373), long sessions still hit a second jank when the entire older history arrived as one full-document refresh. This change backfills that history in ack-paced slices instead.

- The Host splits the deferred prefix into `conversation-viewer-history-chunk` slices (nearest the visible tail first, aligned to interaction-group boundaries), sending each only after the previous slice's correlated receipt. Each slice carries the cumulative content signature, so once the oldest slice lands the Webview's applied signature equals the full content's and later delta publications still omit HTML.
- The Webview prepends each slice directly below the `Loading earlier messages…` placeholder, compensates the scroll offset so the reader's viewport is stable, updates per-message signature bookkeeping for later reconciles, and clears the loading notice on the completing slice. Mermaid decoration runs once on the completing slice instead of once per slice.
- Deferred auxiliary state (comments / project comments / bookmarks) no longer supersedes an in-flight backfill with a full-document refresh; it rides the completion publication, which is HTML-free when the content is unchanged.

Safety properties:

- Stale slices (lower request id, wrong generation, no placeholder) are dropped without acknowledgement; a stale *receipt* can never cancel an in-flight navigation or refresh because the plan is anchored to the partial publication's request id.
- Any page publication supersedes the plan; a lost or unacknowledged slice falls back to a single full-content refresh, still guarded by the generation-level incomplete-content watchdog. The watchdog now also recovers a lost completion publication even when its signature matches the applied content, so deferred auxiliary state cannot strand.
- Previous-generation Webview scripts ignore the new message type and are healed by the same watchdog; the byte-exact compatibility fixture stays green with explicit strips for the new code.

## Verification

- `npm run test-compile`
- `node --test tests/integration/dashboard/conversationViewer.test.js` — 127/127, including new coverage: ack pacing and slice ranges, group-boundary integrity (including a group larger than one slice), the 48-message single-refresh boundary, chunk watchdog recovery, auxiliary-state deferral and early-restore supersede, stale-receipt anchor race against an in-flight refresh, and lost-completion recovery
- `node --test tests/browser/conversationViewer.test.js` — 162/162, including in-place prepend order, scroll compensation (with non-vacuous preconditions), stale slice drops, completion status/signature settle, and HTML-free delta after chunks
- `node scripts/run-dashboard-webview-checks.js`, `scripts/run-conversation-performance-checks.js`, `npm run test:behavior-contracts`, `npm run lint` (only pre-existing warnings), architecture guards, `git diff --check`
- Three independent read-only review passes plus a final integration re-review; all P1 findings were fixed and pinned by tests, final pass reports clean

## Skill harvest

No skill change. The one workflow wrinkle — the byte-exact previous-version fixture derivation broke because a legacy strip anchored on adjacency to `function postNavigation`, which the new functions displaced — is resolved in the test itself (the new strips run before the adjacency-anchored one), and the pinned SHA makes any future drift fail loudly.

## Owner approvals

```
approve 4b0378165b040475b4305b258a78f3527b9a5955
```

### DIFF
--- a/media/conversationViewerScripts.js
+++ b/media/conversationViewerScripts.js
@@ -406,6 +406,9 @@
         renderGeneration: 0,
         appliedHtmlSignature: undefined,
     };
+    // The status text without the deferred-history notice, so a completing
+    // backfill chunk can clear that notice without a full page message.
+    var baseTranscriptStatus = '';
     var readingAnchorController =
         window.__agentPivotConversation.readingAnchor.create({
             scroll: scroll,
@@ -1893,6 +1896,7 @@
         if (message.partial) {
             statusMessages.push('Partial history — showing newest inputs.');
         }
+        baseTranscriptStatus = statusMessages.join(' ');
         if (messages.querySelector('.conversation-deferred-messages')) {
             statusMessages.push('Loading earlier messages.');
         }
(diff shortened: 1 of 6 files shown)


hzcheng:
approve 4b0378165b040475b4305b258a78f3527b9a5955

PR #375, opened 2026-08-25
Add Hand off action that starts a relay chat from an existing AI session

hzcheng:
## Summary

- Session rows (history and active) replace the inline Pin button with a **Hand off** button; Pin moves exclusively into the overflow context menu, and Hand off is also available from that menu so it stays reachable at narrow widths.
- Hand off reuses the existing provider/profile preset dropdown: the picked target starts a fresh chat in the source session's worktree (or its workspace root on multi-root setups) whose first-turn prompt names the source provider, session id, working directory, and — when the provider exposes it — the on-disk transcript path, so the new chat reads the prior context and continues the task.
- An optional `initialPrompt` is plumbed through the new-session launch specs of Codex, Kimi, and Claude (Claude's builder gains a prompt argument), and the `handoff-ai-session` webview message goes through the same profile whitelist validation as quick-create plus bounded `projectId`/`sourceSessionId` checks.

## Safety

- Handoff fails closed when the source session's worktree is unavailable (PRD §6.4) instead of falling back to a default directory scope.
- The handoff prompt frames the source transcript and session metadata as untrusted historical context and tells the new chat not to act on instructions found inside them without user confirmation.
- Contract: `SESSION-HANDOFF-001` in `docs/testing/behavior-contracts.json`.

## Verification

- `npm run test:compile` (tsc + dashboard bundle) — clean
- `npm run test:deterministic:run` — 1598 tests, 0 failures
- `node scripts/run-dashboard-webview-checks.js`, `run-ai-session-safety-checks.js`, `check-behavior-contracts.js` — all exit 0
- Browser: `quickSessionCreate`, `worktreeGroups`, `activeSessionCardInteraction` — 125 tests, 0 failures, including two new tests covering both handoff entry points and non-leakage of handoff context into ordinary quick-create
- Multi-angle read-only review (host correctness, webview/UX, security, repo conventions) with all blocker/major findings fixed in a follow-up commit
- `SKIP_NPM_CI=1 npm run install-local` — both VSIX installed into the active VS Code Server with byte-level hash verification of `dist/dashboard.js`, `media/webviewDashboardBundle.js`, and `media/styles.css`

## Skill harvest

No skill change justified. The only process failure this task (a masked failing exit code from piping `check-behavior-contracts.js` through `grep`) is already covered by `review-fix-commit-loop`'s "judge gate scripts by their own exit code" rule — a compliance gap, not an instruction gap.

## Owner approvals

```
approve 05cb1cb10d1327e9ecb6af056acf37c97ab12533
```


hzcheng:
approve 05cb1cb10d1327e9ecb6af056acf37c97ab12533

PR #378, opened 2026-08-25
perf: patch only the trailing interaction while a conversation streams

hzcheng:
## Summary

This PR makes Conversation switching and history loading noticeably smoother:

- Patch only the streaming tail when it is safe, with capability/version-skew fallbacks to full HTML.
- Keep deferred history backfill ack-paced and preserve reading position while it completes.
- Load older retained-window pages only after the reader reaches the top, rather than blocking open or session switching.
- Correlate every earlier-page request and explicitly settle busy, unavailable, no-progress, and timed-out reads. A timed-out read is aborted after four seconds, so the Webview never remains permanently on `Loading earlier messages`.
- Avoid progressive rendering for ordinary pages that would otherwise show a loading placeholder and immediately repaint the full transcript; only genuinely large histories use bounded chunks.

## Verification

- `npm run test-compile`
- `node scripts/run-dashboard-webview-checks.js`
- `node --test tests/unit/aiSessions/conversationViewerProtocol.test.js`
- `node --test tests/integration/dashboard/conversationViewer.test.js` (136 passing)
- `node --test tests/browser/conversationViewer.test.js` (passing)
- `git diff --check`

## Skill harvest

No skill change. The implementation follows the existing resilient Webview mutation protocol: every asynchronous Webview request now has a correlated terminal result, and the previous-script fixture continues to guard version skew.

## Owner approvals

```
approve 1973c428ca841ad2f2fd546672c2c0dea62e793a
```


hzcheng:
approve 1973c428ca841ad2f2fd546672c2c0dea62e793a

PR #385, opened 2026-08-26
perf: skip duplicate conversation target loads

hzcheng:
## Summary

- keep an already-authoritative conversation document instead of rereading or rerendering an exact duplicate target
- short-circuit the dashboard follow path before snapshot resolution when it returns to the current session
- cancel a stale preflight/read on an A → B → A reversal and keep terminal authority bound to A

## Validation

- `npm run test-compile`
- `node --test tests/integration/dashboard/conversationOpenLatest.test.js`
- `node --test tests/integration/dashboard/conversationViewer.test.js`
- `npm run test:coverage && node scripts/check-changed-coverage.js`
- `git diff --check`

## Skill harvest

No skill change: this was a product-path optimization with no reusable agent workflow gap beyond the already-applied review and publishing guidance.

## Owner approvals

```
approve dab74545e0163053192a8457839e8a558648ab7e
```

### DIFF
--- a/src/aiSessions/conversation/composition.ts
+++ b/src/aiSessions/conversation/composition.ts
@@ -547,6 +547,19 @@ function createAvailableConversationCapability(
                 return 'closed';
             }
             const intent = beginViewerIntent();
+            const currentTarget = viewer.getCurrentTarget();
+            if (currentTarget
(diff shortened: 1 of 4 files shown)


hzcheng:
approve dab74545e0163053192a8457839e8a558648ab7e

PR #392, opened 2026-08-26
ci: shard the Linux quality gate into four parallel jobs

hzcheng:
## Summary

Split the serial ~6-minute `quality-linux` gate in `verify.yml` into four parallel shard jobs so wall-clock time approaches the slowest shard instead of the sum of all checks. Local timing showed the work is heavily concentrated (coverage suite ≈51%, browser tests ≈30%), so the shards are:

- `linux-core` — compile, lint, coverage run + both coverage gates (keeps `fetch-depth: 0` and `COVERAGE_DIFF_BASE`)
- `linux-browser` — compile, Chromium browser tests, dashboard checks, conversation performance (only shard that installs Chromium)
- `linux-safety` — compile, remote conversation sources, safety/tmux logic, architecture rules
- `linux-release` — compile, brand verification, behavior contracts, release notes, release packaging (owns the packaging output)

The required `quality-linux` check stays as a cheap aggregate job with `if: always()` that fails explicitly on any non-success shard — a `needs`-skipped job would otherwise report success to branch protection. Branch protection does not need to change.

Drift guard: `test:ci:linux` remains the serial source of truth for local runs, and the new `validateLinuxShardScripts` contract (wired into `validateQualityGateScripts`) requires the four shards to partition it exactly, so a newly added gate cannot silently drop out of CI. The aggregate reads the `needs` context via the unit-tested `scripts/check-ci-shard-results.js`.

Also fixes `.vscodeignore` listing `.worktree/**` while the real directory is `.worktrees/` — local packaging previously swept every worktree checkout into the VSIX (110k files / 485 MB, OOM-killing the packaging checks); now 91 files / 1.9 MB.

## Verification

- All four shard scripts executed locally end-to-end, including `test:release-packaging` with real VSIX inspection (it re-validates the new workflow structure)
- Changed-line coverage gate: 95.98% (167/174) against `origin/main`, threshold 80%
- Focused suites green: `ciContracts`, `ciShardResults`, `changedCoverage`, `prBody` (60 tests)
- Aggregate script smoke-tested manually for success / failure / skipped shard results
- `git diff --check` clean

## Skill harvest

No skill change. The two friction points in this task were (a) the `.vscodeignore` gap, a product bug fixed in this PR and pinned by `run-release-packaging-checks.js`, and (b) a workflow-mutation test anchor miss, already covered by the existing `ciContracts` contract tests. Existing worktree/review-loop skills applied cleanly.

## Owner approvals

```
approve 053bf699663467fa0612aef21cb9a2c05209b7c1
```


hzcheng:
approve 053bf699663467fa0612aef21cb9a2c05209b7c1

PR #397, opened 2026-08-26
ci: auto-publish tagged minor/major releases to the VS Code Marketplace

hzcheng:
## Summary

Add a `publish-marketplace` job to `release-vsix.yml` that auto-publishes tagged releases to the VS Code Marketplace:

- Runs after the GitHub release job and publishes the **exact VSIX artifacts** produced by that job (downloaded via workflow artifact), UI Bridge before the main extension, authenticated with the `VSCE_PAT` repository secret.
- Publishes on **minor/major bumps** (counting digits from the right: positions 2 and 3 of `x.y.z`): the decision step compares the previous `v*.*.*` tag's `major.minor` prefix with the released version and skips patch-only bumps, keeping the marketplace on the previous minor line.
- A new `publish_marketplace` workflow_dispatch input forces a publish (e.g. shipping an important patch); when forced without credentials the job fails loudly instead of silently skipping.
- When `VSCE_PAT` is not configured, the job warns and skips so releases keep working until the secret is set up.

The release job now exposes `version`/`tag`/`vsix_file`/`bridge_vsix_file` outputs consumed by the publish job.

Contract coverage: `validateReleaseWorkflow` in both `scripts/lib/ciContracts.js` and `scripts/run-release-packaging-checks.js` pins the new job (needs/permissions/runner, full-history checkout, tag comparison, artifact download, VSCE_PAT auth, bridge-before-main ordering, release job outputs), with mutation tests in `tests/unit/tooling/ciContracts.test.js`. Registered as RELEASE-MARKETPLACE-PUBLISH-001 in the behavior catalog.

## Required setup after merge

Create an Azure DevOps PAT with the **Marketplace > Manage** scope (all accessible organizations) for publisher `hzcheng` and store it as the `VSCE_PAT` Actions secret. Until then the job warns and skips.

## Verification

- `npm run test-compile`
- `node --test tests/unit/tooling/ciContracts.test.js` (35 pass, incl. 2 new RELEASE-MARKETPLACE-PUBLISH-001 tests)
- `npm run test:release-notes`
- `npm run test:release-packaging`
- `npm run lint`
- `git diff --check`

## Skill harvest

No skill change justified. One friction point: the release workflow contract is duplicated between `scripts/lib/ciContracts.js` and `scripts/run-release-packaging-checks.js`, so the new job initially failed the second validator; that is a code-duplication observation about this repo's release checks, not a reusable agent-workflow rule missing from an existing skill.

## Owner approvals

```
approve 7d8cab9802edb61269b2824aca04b8f012bc9b43
```


hzcheng:
approve 7d8cab9802edb61269b2824aca04b8f012bc9b43

PR #399, opened 2026-08-27
perf: progressively render heavy conversations

hzcheng:
## Summary

- Render the newest complete interaction groups first for byte-heavy conversation pages, then backfill older groups after Webview acknowledgements.
- Keep interaction groups atomic so worklog structure and incremental signatures remain correct.
- Add delivered-payload timing fields to distinguish full HTML, patches, and frame reuse.

## Verification

- npm run test-compile
- node --test tests/integration/dashboard/conversationViewer.test.js
- npm run lint
- npm run test:behavior-contracts
- git diff --check

## Skill harvest

No skill change: the existing review, verification, and publication workflows already covered the reusable lessons from this change.

## Owner approvals

approve 2cd3c1096f6c088befc81fb3e94217bda6c38149

PR #402, opened 2026-08-27
fix: converge stranded progressive conversation history

hzcheng:
## Summary

Opening a large conversation renders the recent tail first and backfills older
history in ack-paced slices, correlating each slice against the partial page's
request id. Every path that yields that backfill assumed a superseding load
would publish the full document instead — but `performAuthoritativeLoad` claims
the request counter before it can know whether it will publish, and a refresh
that resolves to "nothing changed" publishes nothing at all.

So a single session-watch invalidation during the first paint — the provider
rewriting its transcript, or any touch that leaves the content identical —
orphaned the in-flight slice's receipt, cancelled the plan, and left the reader
with a `Loading earlier messages` placeholder that the Webview's 8s watchdog
relabelled `Earlier messages are taking longer than expected.` Nothing on the
Host was still armed: no slice, no publication, no timer. Earlier history never
arrived, and the transcript stayed stuck until the user switched sessions.

Reproduced deterministically against the unfixed build with the existing
integration harness, using fixtures that return a byte-identical outline and
page (a pure no-op refresh):

| Timing of one no-op watch tick | Result before this PR |
| --- | --- |
| Between slice send and slice receipt | 1 slice sent, `backfill-stale-anchor`, **zero** page publications, placeholder retained |
| Before the partial page's receipt | **zero** slices planned, no diagnostic at all, placeholder retained |
| No tick (control) | 4 slices, `complete: true` |

No Host timer remained armed in either failing case, so the state was terminal
rather than slow.

Two changes:

1. **Hand the request counter back when a load publishes nothing.**
   `releaseUnpublishedRequestId` restores the counter on the "nothing changed"
   short-circuit, but only while the counter is still the one that load claimed
   *and* `latestPublication` is exactly the publication being restored to — so
   it can never revive a load that a newer delivery already superseded.
2. **Give the open incomplete-content obligation the watchdog its callers
   already assumed.** The code comments referenced a "generation-level
   incomplete-content watchdog" that did not exist as an independent timer.
   Now any yield path arms it (`publishDeferredMessages` early return,
   `sendNextProgressiveBackfillChunk` suspended/anchor mismatch,
   `acknowledgeHistoryChunk` stale anchor, `recoverProgressiveBackfill`
   yield/suspended), any pending step clears it (`recordPublicationDelivery`,
   `scheduleProgressiveBackfillTimeout`), and on expiry one full-content
   refresh retires the placeholder. This covers the residual yield paths that
   fix 1 does not — for example a page read that stalls forever.

Also registers `CONVERSATION-LARGE-SESSION-PERFORMANCE-002` in
`docs/testing/behavior-contracts.json`. It was already cited by two owner test
files from #399/#400 but had no catalog entry, so the behavior was owned by
nothing.

Deliberately **not** included: deferring the session-watch `refresh()` behind an
open backfill (the way `revalidateLatest` already does). Once the counter stops
leaking it is unnecessary, and it would delay live streaming updates behind the
backfill. A genuine content change taking over the anchor and publishing the
full page is the existing design intent.

## Owner approvals

```
approve 57bcf8291b81740bc64c1a48c5c6afe6af444168
```

## Skill harvest

updated .skills/fixing-regressions-with-ci — `npm run test:behavior-contracts`
validates catalog→owner only (confirmed in `scripts/check-behavior-contracts.js`
and its own header comment), so `CONVERSATION-LARGE-SESSION-PERFORMANCE-002`
reached two owner test files with no catalog entry and no owner while every
required check stayed green; the skill now requires confirming the reverse
direction with `grep -c '<BEHAVIOR-ID>' docs/testing/behavior-contracts.json`.
Validated with `node scripts/run-skill-management-checks.js`,
`tests/contract/skills/*.test.js` (40/40), and
`tests/unit/tooling/repositorySkills.test.js` (10/10).

## Verification

RED first, against the unfixed implementation — each new test failed for the
reported regression:

- `an unchanged refresh must not strand the deferred history`
- `the partial receipt must still plan the deferred backfill`
- `an open incomplete-content obligation must be watched`

(`linux-core`); `tests/browser/conversationViewer.test.js` →
`test:ci:linux:browser` (`linux-browser`). Both are `pull_request` checks, so
PR-head files govern them.

All gates below were re-run fresh after rebasing onto `71b0ef4e`:

| Check | Result |
| --- | --- |
| `LARGE-SESSION-PERFORMANCE-002` focused (28 tests, 25 pre-existing) | 28/28 |
| `npm run test:deterministic:run` | unit 1348, contract 1195, integration 469 — all pass |
| `npm run test:browser:run` | 424/424 |
| `npm run test:conversation-performance` | pass, within budget |
| `npm run test:behavior-contracts` | pass, 11 release blockers |
| coverage baseline + changed-coverage | pass, 86.01% (123/143) vs `71b0ef4e` |

The browser assertions were added where the Webview code already behaved
correctly, so their mutation sensitivity was proven explicitly: injecting
"a stalled page refuses every later delivery" and "the stalled notice sticks
past its recovery" each failed the new assertions; both were reverted and the
test reconfirmed green with an empty `git diff`.

### DIFF
--- a/.skills/fixing-regressions-with-ci/SKILL.md
+++ b/.skills/fixing-regressions-with-ci/SKILL.md
@@ -62,6 +62,13 @@ Turn every confirmed regression into a CI-owned behavior before changing product
      red run as green. Use `set -o pipefail` or read the stored log.
 7. **Keep the behavior catalog current**
    - In every automated owner file, use literal behavior IDs for the behaviors it owns. After any implementation, owner, or catalog change, run `npm run test:behavior-contracts`.
+   - That command validates catalog→owner only: every entry's owner files
+     exist and cite its ID. It never reads a test for IDs the catalog is
+     missing, so a behavior ID that appears in a test but in no catalog entry
+     passes every required check silently and the behavior ends up owned by
+     nothing. When putting an ID in a test, confirm the reverse direction
+     explicitly: `grep -c '<BEHAVIOR-ID>' docs/testing/behavior-contracts.json`
+     must be non-zero.
    - Classify commits by changed paths and protected behavior, never by a subject prefix. Treat `.skills/` and skill-owner tests as implementation paths.
    - In repositories with both `origin` and `upstream`, pass the intended `--repo <owner/name>` to every `gh pr checks`, `gh run view`, and Actions log command. Do not let local remote ordering select the repository.
 
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -4772,6 +4772,21 @@
     ]
   },
   {
+    "id": "CONVERSATION-LARGE-SESSION-PERFORMANCE-002",
+    "domain": "webview",
+    "title": "Progressively rendered large AI Conversations always converge on their deferred history instead of stranding a loading placeholder",
+    "priority": "P0",
+    "status": "automated",
+    "owners": [
+      "tests/browser/conversationViewer.test.js",
+      "tests/integration/dashboard/conversationViewer.test.js"
+    ],
+    "evidence": [
+      "src/aiSessions/conversation/viewer.ts",
+      "src/webview/conversationViewerScripts.js"
+    ]
+  },
+  {
     "id": "CONVERSATION-VIEWER-USER-EMPHASIS-001",
     "domain": "webview",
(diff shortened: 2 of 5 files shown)


hzcheng:
approve 57bcf8291b81740bc64c1a48c5c6afe6af444168

PR #403, opened 2026-08-27
perf: stop auxiliary state from resending switched conversations

hzcheng:
## Summary

Switching conversations paints the recent tail first and backfills older
history in ack-paced slices. Auxiliary state — restored comments, bookmarks,
and the subagent list — was published as an ordinary page. A fresh
publication's content signature can never match what the Webview already
applied, so that page carried the **entire transcript**.

Worse, it also claimed the request counter, so the partial page's receipt then
failed `isCurrentPublication` and the slice-paced backfill never started at
all. Any switch into a session that has a subagent — or whose comment restore
won the race — therefore discarded the progressive first paint and
re-sanitized, re-parsed, and re-laid-out the whole history in one blocking
Webview task.

Confirmed from a user's real diagnostics: every `follow` event was immediately
followed by a non-progressive `refresh` of 338–695KB, with observed Webview
application times of 150–2671ms (and 10019ms on the previous build). Isolated
in the harness — same 100-interaction session, the only difference being one
subagent:

```
no subagents:  first paint 11KB -> 1 history chunk       -> backfill runs
one subagent:  first paint 11KB -> 0 history chunks      -> refresh html=91KB
```

### Fix

Auxiliary state is not content, so the subagent list now travels the same
deferred `pendingRestoredAuxiliaryState` path that comments and bookmarks
already use, and that path is held back while a deferred-history obligation is
open. It lands once as a state-only envelope with the HTML omitted.

Measured on a 120-interaction session with one subagent and restored comments:

| | before | after |
| --- | --- | --- |
| first paint (blocking) | 19KB | 19KB |
| **largest single Webview task** | **189KB** | **38KB** |
| page publications carrying HTML | 1 (189KB) | **0** |
| history slices | 0 | 5 (170KB, ack-paced) |
| subagents + comments delivered | yes | yes (0-byte envelope) |

Total content is unchanged — the win is that the largest blocking parse drops
5x and the rest arrives in slices that yield between each, so the transcript
stays readable and scrollable throughout the switch.

### Changed expectations (deliberate)

- One test asserted `the restored auxiliary state rides a full-content
  refresh, which completes the progressive page directly instead of chunking`
  with `history-chunk` count `0`. That froze an accidental performance
  behavior; it now asserts the opposite and is renamed onto
  `CONVERSATION-LARGE-SESSION-PERFORMANCE-003`.
- Four subagent tests modelled a Webview that never acknowledges any
  publication, which no real Webview does. A state-only envelope requires an
  applied receipt (`A state-only envelope must never update an outgoing or
  unready document`), so they now send that receipt. Their supersession and
  in-place-switch assertions are unchanged.

Registers `CONVERSATION-LARGE-SESSION-PERFORMANCE-003`.

## Owner approvals

```
approve 70f1b99a4029d10ff655985c433b0fa916c531b1
```

## Skill harvest

updated .skills/protecting-main-with-worktrees — two preventable failures on
this branch: an absolute path rebuilt without the `/.worktrees/<topic>` infix
landed a skill edit in the primary checkout and reported success; later the
worktree vanished mid-task and `npm run test-compile` from the emptied path
walked up to the primary `package.json` and cleaned/rebuilt the primary
`out/`. The existing guardrail only covered inspecting after the fact, so the
skill now records how to avoid both. Validated with
`node scripts/run-skill-management-checks.js`,
`tests/contract/skills/*.test.js` (40/40), and
`tests/unit/tooling/repositorySkills.test.js` (10/10).

## Verification

RED first, against the unfixed implementation:

- `discovering subagents must not resend the conversation` (actual 1, expected 0)
- `the auxiliary restore must not supersede the progressive page` (actual 1, expected 0)

(`linux-core`); `tests/browser/conversationViewer.test.js` →
`test:ci:linux:browser` (`linux-browser`). Both are `pull_request` checks.

| Check | Result |
| --- | --- |
| `LARGE-SESSION-PERFORMANCE` focused (61 tests) | 61/61 |
| `npm run test:deterministic:run` | unit 1348, contract 1195, integration 470 — all pass |
| `npm run test:browser:run` | 425/425 |
| `npm run test:conversation-performance` | pass, within budget |
| `npm run test:dashboard:run`, `test:safety:run` | pass |
| `npm run test:behavior-contracts` | pass, 11 release blockers |
| coverage baseline + changed-coverage | pass, 100.00% (18/18) vs `bf75713f` |

The new browser assertion guards already-working Webview code, so its mutation
sensitivity was proven: making the Webview ignore `subagents` when `html` is
omitted failed it with `the panel must list the discovered subagent`; the
script was restored and the test reconfirmed green with an empty `git diff`.

### DIFF
--- a/.skills/protecting-main-with-worktrees/SKILL.md
+++ b/.skills/protecting-main-with-worktrees/SKILL.md
@@ -57,3 +57,17 @@ job.
 - Never assume `origin/main` when the repo also has `upstream` or the user named a different target.
 - If a feature branch tracks a deleted remote after merge, that is expected; remove the worktree after checking it is clean.
 - If a command accidentally affects the primary checkout, stop and inspect before continuing.
+- An absolute path is the main way edits leak into the primary checkout: the
+  two trees differ only by the `/.worktrees/<topic>` infix, so a path
+  reconstructed from memory silently targets the primary checkout and the edit
+  appears to succeed. Build every absolute path from the worktree root, and
+  after editing check `git -C <primary> status --short` for files you did not
+  intend to touch.
+- A worktree can vanish underneath a session (a parallel agent, a cleanup, a
+  removed directory). `npm run` from the emptied path then walks up to the
+  primary checkout's `package.json` and silently operates on it — a build
+  script such as `test-compile` will clean and rebuild the primary `out/`.
+  Before running builds or tests, confirm the cwd is still a live worktree
+  (`git rev-parse --show-toplevel` matches it, and it appears in
+  `git worktree list`); if it does not, re-create the worktree instead of
+  running anything from that path.
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -4787,6 +4787,21 @@
     ]
   },
   {
+    "id": "CONVERSATION-LARGE-SESSION-PERFORMANCE-003",
+    "domain": "webview",
+    "title": "Auxiliary AI Conversation state never discards a progressive render or resends unchanged conversation content",
+    "priority": "P0",
+    "status": "automated",
+    "owners": [
+      "tests/browser/conversationViewer.test.js",
+      "tests/integration/dashboard/conversationViewer.test.js"
+    ],
+    "evidence": [
+      "src/aiSessions/conversation/viewer.ts",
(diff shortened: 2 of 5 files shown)


hzcheng:
approve 70f1b99a4029d10ff655985c433b0fa916c531b1

PR #405, opened 2026-08-27
perf: restore large conversations from their stashed Webview frame

hzcheng:
## Summary

Switching away from a conversation detaches its DOM into the Webview's frame
cache so that returning can reattach it for nothing: no HTML on the wire, no
sanitize, no parse, no layout. The Host offers that restore only when the
publication's content signature equals the token the Webview reported for the
frame — and for any conversation large enough to render progressively, it never
can. The frame holds the **converged** document while an `initial` switch
publishes the **partial** first paint, so the two signatures describe different
content by construction.

Returning to a conversation is the most common switch there is, and the frame
cache exists precisely for it, so every return visit resent the first paint and
replayed the whole backfill. The preflight had already reattached the cached
frame as a visual preview, so the reader watched the conversation appear and
then get torn down and rebuilt behind the deferred-history placeholder.

Isolated with a harness that drives A (100 interactions) → B → A exactly as a
real Webview does, acknowledging each publication and slice:

```
before:  restoreFrame false | 158KB HTML on the wire | 4 slices replayed (139KB) | 7 messages
after:   restoreFrame true  |   0KB HTML             | 0 slices                  | 2 messages
```

### Fix

`restorableFramePublication` checks the reported frame first: a progressive
candidate whose session is on that list renders whole once to see whether the
frame still matches, and delivers that whole publication when it does —
`deliverPublication` then turns it into a `restoreFrame` with the HTML omitted.
A mismatch (the transcript moved on while away) falls back to the progressive
first paint, so the Host pays one extra render and nothing else changes.
`hostSessionSwitchMs` is unchanged at 2.7ms.

It also skips the probe when the reported token equals the live applied
signature, because the ordinary delta path is already free in that case.

Owned by `CONVERSATION-LARGE-SESSION-PERFORMANCE-001`, which already covers
session switching and the frame-restore contract; no catalog change needed.

This is the third slice of the switch-experience work, after #402
(stranded deferred history) and #403 (auxiliary state resending the
transcript). Host-side load time was measured and ruled out as a bottleneck:
`loadMs - applicationMs` on real switches is 6–70ms, and
`hostSessionSwitchMs` is 2.7ms.

## Owner approvals

```
approve db0acbf39de4eb61086e1fde1d72dbed6b0059f9
```

## Skill harvest

no skill change — the only near-miss on this branch was a scratch probe whose
size-threshold assertion (`> 100_000` bytes) would have passed on the observed
91KB defect; the committed tests already assert categorically instead
(`typeof message.html === 'string'` count must be 0, `restoreFrame === true`),
so the lesson is already embodied in the tests rather than being a missing
instruction. No existing skill proved wrong or incomplete here.

## Verification

RED first, against the unfixed implementation:
`a reported frame for unchanged content must be restored` (actual `undefined`,
expected `true`).

(`linux-core`), a `pull_request` check.

`origin/main` released 1.2.0 mid-branch. Because this branch had already been
pushed, it was brought current by merging `origin/main` rather than a
force-push, and every gate below was then re-run on the merged tree:

| Check | Result |
| --- | --- |
| `LARGE-SESSION-PERFORMANCE` focused (62 tests) | 62/62 |
| `npm run test:deterministic:run` | unit 1348, contract 1195, integration 471 — all pass |
| `npm run test:browser:run` | 425/425 |
| `npm run test:conversation-performance` | pass; `hostSessionSwitchMs` 2.689ms |
| `npm run test:dashboard:run`, `test:safety:run` | pass |
| `npm run test:behavior-contracts` | pass |
| coverage baseline + changed-coverage | pass, 100.00% (44/44) vs `7168db2a` |

### DIFF
--- a/src/aiSessions/conversation/viewer.ts
+++ b/src/aiSessions/conversation/viewer.ts
@@ -2610,7 +2610,13 @@ export class ConversationViewer implements ConversationViewerApi {
             }
             const progressiveCandidate = updateKind === 'initial'
                 && this.canProgressivelyRender();
-            const publication = this.createPublication(
+            const publication = this.restorableFramePublication(
+                requestId,
+                generation,
+                updateKind,
+                progressiveCandidate,
+                target
+            ) || this.createPublication(
                 requestId,
                 generation,
                 updateKind,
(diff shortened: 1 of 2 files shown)


hzcheng:
approve db0acbf39de4eb61086e1fde1d72dbed6b0059f9

PR #408, opened 2026-08-28
feat: enhance controlled conversation rendering

hzcheng:
## Summary

- Render conversation diffs, structured content, safe workspace links, KaTeX, and bounded charts.
- Add controlled reference cards, code line highlighting, sortable tables, and accessible task and chart semantics.
- Stop a bar chart card collapsing onto the chart below it. The card carries the
  type modifier `conversation-chart-bar`, and the `<progress>` element inside it
  carried the same class, so the rule written for the progress element
  (`height: 0.75rem`) also matched the card. The card shrank to one bar's height
  while its rows laid out at full size outside it and painted over the next
  chart. Only bar charts were affected; no rule matches the `-line` or `-pie`
  modifiers. The progress element is now named `conversation-chart-progress`.
- Version every Conversation Viewer static asset per document, so a reloaded Webview cannot reuse stale CSS or scripts after an extension update.
- Harden rich-rendering refresh state, KaTeX CSP/sanitization, and worktree-relative link resolution.

## Verification

- `npm run test-compile`
- `npm run test:ci:linux:safety`
- `git diff --check`

Rendered-surface evidence for the chart fix: with the shipped stylesheet, the
bar chart card measured 24px tall while its body measured 124px and overflowed
85px past the card. After the fix the card measures 137px and contains its body.

## Skill harvest

No skill change. Three consecutive fix attempts missed this defect, but each
miss violated a rule that the applicable skills already state rather than
exposing a missing one.

The overlap guard added in an earlier round asserted that adjacent chart *card
boxes* do not overlap. A collapsed card satisfies that assertion precisely
because its bottom edge sits too high, so the guard passed while the reported
symptom persisted. `fixing-regressions-with-ci` step 4 already requires
observing the new test fail for the reported regression before the production
edit; that guard was never observed failing, and doing so would have exposed
its vacuity immediately. The replacement asserts that every chart card contains
its own body, which no collapsed card can satisfy, and it was confirmed RED
against the reported content before the fix.

Separately, `npm run install-local` exits 0 when its internal `npm ci` step
hard-fails under a worktree (`EALLOWSCRIPTS`), so it reports success having
installed nothing. `installing-vscode-extensions-locally` already directs
verifying installed bytes rather than exit status, so this is a defect to
repair in the script rather than a skill gap, and it is out of scope for this
PR.

## Owner approvals

```
approve 2fc27a56905210fd59bbe5382a7aae20eff90395
```


hzcheng:
approve cca43727c139d605a4276ee7c2fcb6e7085eee6a

hzcheng:
approve 2fc27a56905210fd59bbe5382a7aae20eff90395

PR #411, opened 2026-08-28
fix: install local extensions through the worktree dependency guard

hzcheng:
## Summary

- Prepare dependencies in `scripts/build-test-package-install.sh` through the
  worktree dependency guard instead of a bare `npm ci`.

Every line of work runs in a worktree, but the local install script ran
`npm ci` directly. Under an inherited user-level allow-scripts config that
fails outright — `EALLOWSCRIPTS`, because project-scoped installs reject
`--allow-scripts` — so `npm run install-local` could not complete in a worktree
without setting `SKIP_NPM_CI=1` and preparing dependencies separately. A bare
`npm ci` also deletes `node_modules`, which the guard exists to prevent while
another command in the same worktree is compiling or testing.

`scripts/with-worktree-dependencies.js` already handles all three concerns: it
strips the inherited allow-scripts config, holds the per-worktree lock, and
skips the install when dependencies are current.

## Verification

- `npm run lint:ci`
- `git diff --check`
- End-to-end: `npm run install-local` in a fresh worktree, with no
  `SKIP_NPM_CI` workaround, exits 0 and installs both extensions with verified
  bytes. Before this change the same command exited 1 at the dependency step.

The new owner runs the script with recording shims for `npm` and `node` rather
than pattern-matching its source, so the real invocation sequence is asserted.

## Skill harvest

No skill change. This PR corrects a claim I made while shipping #408: I reported
that `install-local` exits 0 when its dependency step fails. It does not — it
exits 1, and a new owner test now pins that. The wrong reading came from
inspecting `$?` after a compound shell command, where it reported the exit
status of the trailing `tail`, not of npm.

`verification-before-completion` already requires confirming command output
before asserting an outcome, so this was a compliance gap rather than a missing
rule. The durable protection is the executable owner added here, which pins the
real exit-code behaviour instead of relying on a one-off shell reading.

## Owner approvals

```
approve 83fdd16b6f697d6c5bd98ac43c54b6db01758d71
```

### DIFF
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -2394,6 +2394,20 @@
     ]
   },
   {
+    "id": "LOCAL-INSTALL-WORKTREE-DEPENDENCIES-001",
+    "domain": "release",
+    "title": "Local install prepares dependencies through the worktree guard",
+    "priority": "P2",
+    "status": "automated",
+    "owners": [
+      "tests/unit/tooling/buildTestPackageInstall.test.js"
+    ],
+    "evidence": [
+      "scripts/build-test-package-install.sh",
+      "scripts/with-worktree-dependencies.js"
+    ]
+  },
+  {
     "id": "OPEN-BRIDGE-CLIENT-001",
     "domain": "open-project",
     "title": "Bridge Client behavior",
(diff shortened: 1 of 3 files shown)


hzcheng:
approve 83fdd16b6f697d6c5bd98ac43c54b6db01758d71

PR #413, opened 2026-08-29
fix: wrap long diff lines instead of overlapping the opposite pane

hzcheng:
## Summary

- Wrap long diff lines so a line wider than its pane stops painting across the
  opposite pane in the side-by-side view.

Each diff row is a two-column grid capped at half the scrollport, but the line
text was `white-space: pre` with no overflow containment. A line longer than
half the width kept its full painted width and rendered straight across the
other pane, so a real Java diff showed both panes' text on top of each other.

Two alternatives were measured and rejected. Clipping the cell stops the overlap
but silently truncates the code. Sizing the panes to their widest line and
handing the overflow to the existing hunk scrollport also stops the overlap, but
the panes then land at a different offset in every hunk and reaching the new
pane scrolls the old one out of view — harder to read than the defect it
replaced.

| Candidate | Overlap | Text clipped | Pane offsets across all hunks | Horizontal scroll |
| --- | --- | --- | --- | --- |
| Wrap (this PR) | none | no | 1 | no |
| Shared content-sized tracks | none | no | 2 | yes |
| Per-pane scroll | 236px | yes | 1 | no |

## Verification

- `npm run test:ci:linux:safety` (includes skill management checks)
- `npm run lint:ci`
- `git diff --check`

Mutation sensitivity, run because the first attempt at this repair was wrong:
reintroducing the original `white-space: pre` fails `no diff line may paint
across into the opposite pane`; reintroducing the rejected shared-track layout
fails `every row in every hunk starts its new pane at the same offset`. The
owner's fixture carries two hunks with different longest lines, because a
single-hunk fixture cannot observe that alignment — which is why the previous
guard missed it.

## Skill harvest

Updated `.skills/fixing-regressions-with-ci`. The first repair in this branch
satisfied its guard and still shipped a worse experience: the guard asserted
only the absence of the reported overlap, so a layout that removed the overlap
while misaligning the panes and forcing a scroll passed cleanly. No existing
rule covered choosing between layout strategies.

Two rules added. A repair that changes a layout strategy must enumerate the
properties a reader depends on — alignment across every repeated group, nothing
clipped, no scroll that hides the content being compared against — assert them
together, and measure each candidate against that set with the real reported
content rendered and looked at. And mutation sensitivity now extends to rejected
candidates, so a guard cannot be satisfied by the alternative that was
discarded.

Validated with `tests/unit/tooling/repositorySkills.test.js` and the safety
shard's skill management checks.

## Owner approvals

```
approve 6470e157dda44d62e40ff3455848ac1c84457216
```


hzcheng:
approve 6470e157dda44d62e40ff3455848ac1c84457216

PR #414, opened 2026-08-29
feat: let readers switch diff line wrapping per file

hzcheng:
## Summary

- Default side-by-side diff lines to unwrapped, clipped at their own pane with
  an ellipsis, and add a per-file header toggle that turns wrapping on.

This matches the VS Code diff editor, whose `diffEditor.wordWrap` defaults to
`inherit` over an editor default of `off`: a change within a line is harder to
compare once the line is broken across visual rows. Clipping at the line's own
edge means a line wider than its pane can never paint across into the opposite
one, and the ellipsis makes the truncation visible rather than silent.

The Webview owns the toggle's pressed state, keyed like the existing diff and
table controls, so a Host-authored refresh restores it.

## Synchronised horizontal scrolling: measured and deliberately excluded

VS Code mirrors `scrollLeft` between two whole-column editors
(`diffEditorViewZones.ts`, bidirectional with a re-entrancy guard). These diffs
are row-major — every row is its own scrollport — so the same approach was
implemented and then measured against the reported content:

| Row | Requested offset | Reached | Max |
| --- | --- | --- | --- |
| `// 初始化Connector` | 200 | 0 | 0 |
| `logger.info(...)` (long) | 200 | 200 | 249 |
| `connector.init(...)` | 200 | 21 | 21 |
| `+ connector.start()` | 200 | 0 | 0 |

Rows whose line is short or absent clamp below the shared offset, leaving the
text ragged, and the line number scrolls out of view with the text because it
sits inside the scrollport. Matching VS Code would mean translating every line
from one file-level offset behind a proxy scrollbar — a larger change than this
one, and out of scope here.

## Verification

- `npm run test:ci:linux:safety` (includes the Webview manifest closed-world check)
- `npm run lint:ci`
- `git diff --check`

All re-run after rebasing onto `origin/main`, with the local install rebuilt and
byte-verified from the rebased tree.

Mutation sensitivity: removing the clip fails `no diff line may paint across
into the opposite pane`; removing the refresh restore fails `must not silently
discard the reader wrap choice`.

Also fixed: both diff header controls resolve to the same file control key, so
focus restore now compares the control kind. Without it a refresh moved focus
from the wrap toggle onto the context toggle. The Viewer fixture hash in
`CONVERSATION-OUTLINE-NAVIGATION-001` moves with the changed Webview script;
that fixture is derived from the live script, so the pin is a tripwire to
update, not a check to bypass.

Unrelated observation: `WEBVIEW-PROJECTS-PANEL-SCROLL-001` failed once under the
concurrent browser shard and passes in isolation. Not touched here.

## Skill harvest

Updated `.skills/fixing-regressions-with-ci`. Both rules involved were added by
the previous PR and both were followed in letter while still missing.

The layout rule was scoped to a *repair*, so designing a new feature escaped it.
A per-row scrollport mechanism was committed to in a plan on DOM reasoning
alone, and only measurement after it was built showed the clamping and the
scrolled-away line numbers above. The trigger now covers any change of layout
strategy, and the evidence is owed before a mechanism is fixed in a plan.

The mutation rule said to restore the implementation but not how. Restoring one
mutated file with `git checkout` reset it to the index and destroyed the
uncommitted feature work under test; it survived only because a build step had
copied it elsewhere. Restoring from a copy taken before the mutation is now
required, and `git checkout`/`git restore` on a file holding uncommitted work is
called out as destructive.

Validated with `tests/unit/tooling/repositorySkills.test.js` and the safety
shard's skill management checks.

## Owner approvals

```
approve cc623a5aeb92e94615eac2ccb01fbbbf1575be69
```


hzcheng:
approve cc623a5aeb92e94615eac2ccb01fbbbf1575be69

PR #415, opened 2026-08-29
fix: keep AI Conversation readers pinned to the tail

hzcheng:
## Problem

Returning to a Session repeatedly landed on a very old interaction, and a
long-running Session drifted away from its newest output — both forcing a
manual scroll to the bottom every time. Two independent defects produced that,
and the second one made the first permanent.

## Defect A — deferred history backfill skipped its scroll compensation

`applyHistoryChunkContent` skipped compensation whenever the reader's offset
was 0, on the theory that offset 0 means "scrolled to the top, waiting for this
history". A progressive first screen shorter than the viewport cannot scroll at
all, so a reader looking at the tail also reports 0. Every backfilled slice then
pushed the tail further out of view and left them on the oldest interaction the
backfill delivered.

Compensate unless the transcript was actually scrollable and the reader was at
its top.

## Defect B — the follow-the-end ResizeObserver watched only the viewport

A transcript growing under a settled reader — lazily decoded images, late
Mermaid renders — moves the end away from them exactly like the viewport
shrinking does, but emits no scroll event. The physical offset silently drifted
past the auto-scroll threshold while the reader never moved, and every later
check reading that offset concluded they had walked away: live refreshes
restored a reading anchor instead of following, and a Session switch stashed a
mid-transcript anchor that each return then faithfully restored.

Observe the transcript as well, so growth beneath a follower brings the end back
to them and the offset stays truthful for those checks.

A first attempt instead overrode the physical check with the tracked follow
intent. That broke two existing `CONVERSATION-READING-FOCUS-001` contracts: a
reader parked 9px from the end must keep their position, and the tracked intent
is stale at message-processing time because the scroll event has not dispatched
yet. Keeping the offset accurate is the correct fix and needed no change to the
stash or refresh logic.

## Verification

- New behavior `CONVERSATION-TAIL-FOLLOW-001` owns all three paths (backfill,
  live refresh, Session switch away and back) in `tests/browser/conversationViewer.test.js`.
- All three tests observed RED first against the unfixed source, failing for the
  reported reason at 110px / 300px / 300px off the tail.
- Mutation-checked each defect independently: reintroducing A fails only the
  backfill test; reintroducing B fails only the refresh and switch-back tests.
- Full `npm run test:ci:linux` green, including the browser suite (447/447 at
  the time of that run), `test:behavior-contracts`, and the integration viewer
  suite (156/156).
- Built and installed locally as `hzcheng.agent-pivot@1.2.1`; both changed
  webview scripts byte-verified against the VSIX in the active Server
  extensions directory.

## Skill harvest

Updated `.skills/resilient-webview-mutation-protocols` (description only). Its
Repository Layout section already owns the rule that `media/` holds
byte-identical copies of `src/webview/*.js` and that both sides must move in one
commit, but the description only triggered on mutation-protocol concerns
(Host-owned mutations, stale acknowledgements, stuck pending UI, focus loss,
partial batch failures). A scroll-position fix in `src/webview/` matches none of
those, so the skill never surfaced, the `media/` mirror was left stale, and the
packaged byte-identity check failed a full verification cycle later. The
description now triggers on the edit itself. Validated with
`tests/unit/tooling/repositorySkills.test.js` (10/10) and
`scripts/run-skill-management-checks.js`.

## Owner approvals

```
approve ca512c642d16a7142dd2a3fd120c452e7b2c0332
```

### DIFF
--- a/.skills/resilient-webview-mutation-protocols/SKILL.md
+++ b/.skills/resilient-webview-mutation-protocols/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: resilient-webview-mutation-protocols
-description: Use when Agent Pivot VS Code Webviews submit Host-owned mutations or authoritative HTML replacement, mirrored persistence, stale acknowledgements, stuck pending UI, focus loss, or partial batch failures are in scope.
+description: Use when editing any Agent Pivot Webview script under `src/webview/` or its `media/` runtime copy, and when Webviews submit Host-owned mutations or authoritative HTML replacement, mirrored persistence, stale acknowledgements, stuck pending UI, focus loss, or partial batch failures are in scope.
 ---
 
 # Resilient Webview Mutation Protocols
--- a/docs/testing/behavior-contracts.json
+++ b/docs/testing/behavior-contracts.json
@@ -1995,6 +1995,20 @@
     ]
   },
   {
+    "id": "CONVERSATION-TAIL-FOLLOW-001",
+    "domain": "webview",
+    "title": "AI Conversation keeps a reader who is following the conversation tail pinned to it across deferred history backfill, live refreshes, and a Session switch away and back, including when lazily decoded images or late diagram renders grow the transcript beneath them without emitting a scroll event",
+    "priority": "P0",
+    "status": "automated",
+    "owners": [
+      "tests/browser/conversationViewer.test.js"
+    ],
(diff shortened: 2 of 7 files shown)


hzcheng:
approve ca512c642d16a7142dd2a3fd120c452e7b2c0332

PR #432, opened 2026-09-06
Retire the pre-catalog project store

hzcheng:
## Summary

Retires the pre-catalog project store now that recovery is complete.

The catalog holds the Machines, Environments and Projects the legacy store used
to hold, so everything that existed to move them across is unreachable. Keeping
it would leave exactly the upgrade machinery this repository decided not to
carry. Removes the activation-time recovery, the import command and its
registration, the conversion module, and the store and controller methods that
applied it, with their tests and the module entrypoint they required.

In their place, activation clears the two settings that backed the old store.
No code path reads them, so leaving them behind only means someone editing
settings.json finds keys that do nothing.

Deliberately unchanged:

- The **local** project store. It still backs the Local Projects that recovery
  reported as left in place. The exported key list is pinned by a test so it
  cannot quietly widen to a setting that still matters.
- The UI Bridge SSH endpoint reader. It is covered by its own tests, needs no
  reinstall, and reading a host's endpoint from the user's own SSH config is
  useful beyond recovery.

## Architecture impact report

Net removal: the Managed Remote store and controller lose their legacy-import
surface, and `MOD-PROJECT-CATALOG` loses the entrypoint that existed only for
it. Module boundaries, closed-world classification, single-writer invariants and
the webview manifest all pass unchanged.

## Owner approvals

```
approve a3551ebdfa650b680e3250c4fa6b0f667a457d54
```

## Skill harvest

no skill change — this branch's lesson was already harvested into
.skills/fixing-regressions-with-ci (read runtime evidence before proposing a
cause); the cleanup itself surfaced no new workflow rule.

## Verification

Run in this worktree at the head above:

- `npm run test:deterministic` — 3259 tests (1460 unit, 1321 contract, 478 integration), 0 failures
- `npm run lint` — exit 0
- `npm run test:safety:run`, `test:architecture-policy`, `test:architecture-guards`
- `npm run test:behavior-contracts`, `test:dashboard:run`, `test:open-projects`
- `node scripts/check-changed-coverage.js` — 100% of changed lines

The cleanup is covered by four tests: it clears a present value at global scope,
writes nothing when the keys are already absent (so repeat activations cannot
churn settings.json), survives a rejected write without aborting startup, and
pins the key list to the two obsolete keys.

Verified in the built artifact: the cleanup ships, the recovery and import
command are gone, and the local project store is untouched.


hzcheng:
approve a3551ebdfa650b680e3250c4fa6b0f667a457d54

PR #434, opened 2026-09-06
feat: save current projects through machine adoption

hzcheng:
## Summary

- make Save Current Project the only project-creation workflow
- adopt an unregistered SSH machine only after a bound confirmation and retain its original alias
- make Window-tab saves visibly pending and idempotent, including across a repeated SSH adoption
- show compact save phases in the Window row and emit correlated host timing diagnostics
- preserve legacy void save callbacks as successful completion while honoring an explicit cancellation
- prevent duplicate managed projects and remove legacy Add Project routes
- rebase onto current main while retaining inline Machine and Project editing flows

## Verification

- npx gulp --production
- npm run test-compile
- focused browser and workspace-save tests (32 passing)
- node scripts/run-open-project-safety-checks.js
- node scripts/run-dashboard-webview-checks.js
- package and verified local Agent Pivot installation bytes
- git diff --check

## Skill harvest

No skill change: the new progress diagnostics followed the existing Webview mutation, regression, review, local-install, and protected-worktree guidance. The only retry was the required rebase after main advanced, which the publishing workflow already prescribes.

## Owner approvals

```
approve f898e959ce7e2c4e008b04a1580bcab96354f188
```


hzcheng:
approve f898e959ce7e2c4e008b04a1580bcab96354f188

PR #443, opened 2026-09-11
Fix saved SSH projects showing the Save button in the current window

hzcheng:
Saved SSH projects could still show the Save button in the current window because the remote Extension Host exposes a file URI without the SSH Machine authority.

Use the current window's own UI Bridge registration for saved-project lookup after checking the workspace kind, environment, and resource path. Apply the resolved identity to the saved state, name, and color while preserving the original workspace identity for session hydration. Add a rendered-HTML regression test based on the observed uquant URI, including peer/stale registrations and catalog removal.

CI follow-up: the existing 50-Machine/500-Project timing check failed twice under concurrent coverage instrumentation (222.6ms and 350.5ms). Move the unchanged 200ms measurement to a standalone required Linux browser-shard step, sharing the same fixture. Retain catalog correctness and payload limits in contract coverage; add gate-wiring and exact-budget rejection tests. Keep the release-packaging command contract synchronized with this required gate. This follow-up changes testing only.

## Validation
- Regression test failed on the original implementation and passes with the fix.
- 608 related workspace and Dashboard contract/integration tests passed (canonical macOS TMPDIR).
- Compile, focused project-matching tests, behavior contracts, lint, Dashboard Webview checks, and release packaging checks passed.
- Local Agent Pivot/UI Bridge and the uquant SSH workspace extension were installed from this build; installed runtime hashes match the VSIX.
- Full Linux core command attempted on macOS: 1494 unit tests passed; two existing workspaceMarkdownFile tests returned stale instead of applied. Both failures were reproduced with origin/main production modules. The subsequent Linux CI core gate passed.

- CI follow-up: 55 focused tests, lint ratchet, instrumented contract/tooling tests, behavior contracts, and the standalone timing gate passed (61.0ms in the final run). Read-only review found no Critical/Important issues.

- Release follow-up: compile, CI tooling tests, and a full fresh VSIX build with release packaging checks passed.

## Skill harvest
No skill change: runtime-identity diagnosis, baseline failure classification, and installed-byte verification are already covered by the existing debugging, review, and installation skills.

## Owner approvals
After validation, the owner can post this exact command as a PR comment:
```
approve c62dfd851847d18f3209eb9b95f3ba0d1d3c57de
```


hzcheng:
approve c62dfd851847d18f3209eb9b95f3ba0d1d3c57de

PR #446, opened 2026-09-19
feat: stream live Codex terminal output in AI Conversation

hzcheng:
## Summary

AI Conversation previously waited until ordinary Codex terminals persisted a completed reply. A shared-daemon observer alone could not see those terminals. Agent Pivot now launches new/resumed Unix Codex terminals with a dedicated terminal-owned app-server and the original remote TUI, letting the conversation viewer receive assistant deltas before completion.

The managed runner preserves working-directory scope, extra workspace roots, initial permission choices, and exact terminal/root binding through new/resume/fork transitions. Ephemeral helpers and viewer subscriptions cannot select the terminal root. Private PID-bound metadata, bounded relay buffers, startup cancellation, and companion cleanup contain lifecycle failures. The packaged runner is included in release checks.

The viewer coalesces updates and reconciles partial text with authoritative completed items. Claude/Kimi transcript watchers expose content when their CLIs persist it.

Follow-up fixes after real-environment testing: a feed attaching after a turn started received a snapshot without the already-started agentMessage item and dropped every later delta (visible as batch-only output when switching back to a chat mid-turn); the method-scoped delta now synthesizes the item. A streaming tail patch replaced the trailing interaction group without the reconcile path's mermaid preservation, re-rendering unchanged diagrams on every delta; figures now move across by source. And a viewer detaching mid-turn lost everything already streamed: the resume snapshot omits the already-started agentMessage item, so the feed retains the last in-flight text per (session, turn) and re-seeds the next attach — a switch away and back no longer regresses visible content.

Earlier: an interrupted Codex turn is served with synthetic per-turn item ids (`item-1`, …) by the companion's in-memory view while the rollout replay exposes the real server ids. The live overlay replaces durable tail content by turn id, so the interaction id flipped underneath the viewer's anchored selection and every later refresh failed closed, republishing unchanged content while deltas kept arriving. The adapter now pins every emitted interaction id per (session, turn, position) — the first emitted id wins and re-applies to durable reloads after the turn leaves the live tail window. Sanitized diagnostics along the feed and viewer refresh gates (the instrumentation that located this) are included.

## Compatibility

- Existing running Codex terminals need one exit/reopen after extension reload; reload alone cannot migrate them.
- `codex app-server` rejects `--profile`, so `-p` launches flatten the profile-v2 file into top-priority `-c` overrides on the companion (a legacy `[profiles.<name>]` table is selected through the `profile` key). Profiles with constructs a flat `-c` list cannot express, missing profile files, unsupported CLIs, and Windows retain ordinary CLI behavior with the notice.
- The managed path was tested with Codex 0.155.0 on reddev-container and 0.155.1 locally. No existing user chat was prompted, interrupted, or reconfigured.

## Validation

- Actual Codex TUI plus production feed/adapter: new, resumed, multi-root, and in-TUI `/new` all produced assistant snapshots before completion. Packaged runner: 286 pre-completion updates; saved final output matched the expected 1091-character reply exactly.
- Focused launcher/binding/builder tests: 51 passed. Incremental browser rendering and release VSIX content checks passed.
- Full Linux regression: 1504 unit, 1396 contract, 509 integration, and 546 browser tests passed. Architecture, behavior, performance, release packaging, and coverage gates passed; changed-line coverage is 93.62% (719/768). The Linux run was resumed at architecture checks after removing macOS `._` archive metadata from the disposable test copy (no production-code change).
- Profile launches: a new contract test drives the packaged runner through a real profile file and asserts the companion receives the flattened provider config while the TUI keeps `-p`; a live `codex app-server` launched with the flattened real `kimi-k3-codewiz` profile serves threads on the profile's model/provider (`codewiz:kimi3` / `codewiz`).
- Id-flip fix: new contract test CONVERSATION-LIVE-001 reproduces the synthetic-id overlay, the pin, and the live-window slide; focused adapter/feed/terminal/coordinator suites pass (159 tests). On the production extension with an interrupted tail turn present, an injected real turn produced 95 growing content refreshes with zero refresh-gate failures; the same chat wedged every refresh before the fix.
- Known clean-base harness limitation: `run-ai-session-safety-checks.js` exits zero without its final banner at its terminal-binding fixture. Its aggregate exit code is not evidence that later assertions in that script ran; focused runtime ownership tests completed explicitly.
- Final read-only integration review found no remaining Critical/Important issues.

## Skill harvest

No skill change: the initial shared-server-only probe failed to follow the existing provider-adapter skill's real-workflow evidence rule. The correction now verifies actual TUI launches, root transitions, production adapters, final text, and the packaged executable. The id-flip investigation reconfirmed the same rule — transport-level probes passed while the real viewer was wedged; only production-extension diagnostics and a direct companion/rollout comparison exposed the synthetic-id dialect. Existing architecture and final-verification rules also already cover the missing public-entrypoint registration caught by the full gate.

## Owner approvals

After reviewing this exact head, the owner can post:

```text
approve 59eb047cc70bddbfd0a89d368160bb614b525e77
```


hzcheng:
approve 59eb047cc70bddbfd0a89d368160bb614b525e77

hzcheng:
approve 59eb047cc70bddbfd0a89d368160bb614b525e77

PR #451, opened 2026-09-21
fix: preserve quoted profile keys in Codex streaming launches

hzcheng:
## Summary

A valid profile containing a quoted project table such as `[projects."/path/to/repo"]` was rejected by the streaming configuration converter. Codex then fell back to ordinary terminal mode, so AI Conversation could not receive live text.

Decode nested quoted keys and emit one inline TOML table per affected root. This preserves paths containing dots, sibling project settings, and configuration from lower layers without changing the user's profile. Unsupported structures continue to use the existing fallback.

## Validation

- Reproduced the original parser failure and ordinary-terminal fallback before the fix.
- 37 focused profile, launcher, and command-builder tests passed, including partial text before completion and exact final text.
- Linux core regression passed: 1512 unit, 1400 contract, and 511 integration tests; changed-line coverage 100% (67/67).
- Behavior contracts, architecture policy, lint, and release packaging checks passed. The macOS full-core attempt exposed three unchanged path/descriptor tests; all three passed in the isolated Linux run.
- Real Codex 0.155.0 on reddev-container with the unchanged `kimi-own` profile selected `kimi-code:k3`, delivered 13 pre-completion updates, and produced exactly the expected 291-character final reply. A separate config/read probe retained unrelated base project trust settings.
- Installed and byte-verified the packaged repair on the local and remote hosts; existing user terminals were not restarted.
- Final read-only review found no Critical/Important issues.

## Skill harvest

No skill change: existing regression guidance already requires the actual failing input, a failing CI-owned test, and real provider verification. This fix follows those rules with the reported quoted project header, launch regression, and isolated remote profile probe.

## Owner approvals

After reviewing this exact head, the owner can post:

```text
approve ad3edaa5750e4fbf62e438c753c7b941d2a55a4e
```


hzcheng:
approve ad3edaa5750e4fbf62e438c753c7b941d2a55a4e

